Issue #5, opened 2018-06-14
Configure Webpack


bernabe9:
Configure Webpack for:
- Compile SCSS code
- Compile Javascript code
- Load images and other assets

stephenp:
@a-axton and I were discussing this. If we're simply compiling / minifying the SCSS into a single file, is a single line with `node-sass` in the package.json file as a npm script sufficient?

bernabe9:
Totally agree, we don't need Webpack to compile scss if we can do it with just one npm script

PR #10, opened 2018-06-14
Compile, minify mastersass and move to dist

stephenp:
Add npm scripts to compile both mastersass and react style libraries.  If we can do this via webpack instead, we should probably move to that.

## Risks
No

## Changes
No visual changes.

## Issue
https://github.com/yankaindustries/mc-components/issues/5

Does not resolve the issue, but addresses some of the concerns. If this isn't the way we want to do it, let's figure out a solution.


package.json
```diff
@@ -14,7 +14,9 @@
     "prebuild": "rimraf lib",
     "build": "npm run build:js && npm run build:scss",
     "build:js": "NODE_ENV=production babel ./src --out-dir ./lib --ignore spec.js,test.js,stories.js",
-    "build:scss": "cpx -p './src/styles/**/*.scss' ./lib/styles",
+    "mastersass-styles": "node-sass --output-style compressed src/styles/index.scss dist/css/mastersass.css",
+    "react-styles": "cpx -p './src/styles/**/*.scss' ./lib/styles",
+    "build:scss": "npm run react-styles ; npm run mastersass-styles",
```
bernabe9: Use `&&` to chain scripts instead of `;`

`"build:scss": "npm run react-styles && npm run mastersass-styles",`

Reference: https://stackoverflow.com/questions/39172536/running-npm-scripts-sequentially

Issue #15, opened 2018-06-18
SCSS copied to lib folder?


stephenp:
@a-axton and I were wondering...why does the scss get copied from src to lib?

bernabe9:
When you publish a library through NPM is a good practice to ignore `src` folder to reduce the module size. Actually this file https://github.com/yankaindustries/mc-components/blob/develop/.npmignore is ignoring the `src` folder.

`lib` or `dist` usually represent the "public" folder

stephenp:
Thoughts on moving all the `lib` contents to `dist`, and then having a `scss` and `css` folder within `styles`? So then anyone can consume the `dist` folder and reference both the sass variables and compiled css if they want it?

stephenp:
https://github.com/yankaindustries/mc-components/pull/18

PR #18, opened 2018-06-18
Moving distributable code to dist folder

stephenp:
- Consolidate JS and CSS code into a single `dist` folder
- Include both `scss` and compiled `css` in `dist/styles` folder

## Risks
No

## Changes
No visual changes

## Issue
https://github.com/yankaindustries/mc-components/issues/15


bernabe9:
should `dist` folder be commited to github???

stephenp:
We have previously committed the dist folder so other members of the org can grab the compiled output (for the css) for any projects they're working on, like static marketing pages or standalone one page apps.  I guess we can switch to just committing src, but they would then have to have some sort of build process for shipping their "apps" which I'm not sure is a valid tradeoff.  Thoughts?

bernabe9:
I see. The problem is that we use Github to distribute our library instead NPM.

With NPM we don't need to commit the `dist` folder, the flow should be:
1- Create PR with changes
2- Merge PR
3- Publish version via NPM

Anyways, I think its a flexible solution to have `dist` folder committed

PR #23, opened 2018-06-19
Move mastersass button styles into c-button component

stephenp:
Instead of updating the component class names (high risk), update the existing class name styles instead (lower risk).

## Risks
Low - Medium

## Changes
Updates how the buttons look based on design-approved comps.  Only updates the primary, secondary, and tertiary button types.

## Issue
N/A

### DIFF
--- a/.storybook/styles.scss
+++ b/.storybook/styles.scss
@@ -8,3 +8,12 @@ body {
   // overwrite mc-sass-lib default body black background
   background-color: transparent !important;
 }
+
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  color: #000;
+}
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,7 +1,3 @@
-// Code that isn't truly approved yet
-// but needs to be implemented to try
-@import "./release-candidate/index.scss";
-
 // External libraries first
 @import "./libraries/libraries";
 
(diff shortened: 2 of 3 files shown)


.storybook/styles.scss
```diff
@@ -8,3 +8,12 @@ body {
   // overwrite mc-sass-lib default body black background
   background-color: transparent !important;
 }
+
```
stephenp: mastersass defines these as white, so it was being inherited. Override for the styleguide.

src/styles/index.scss
```diff
@@ -10,3 +6,7 @@
 @import "./components/components";
 @import "./icons/icons";
 @import "./typography/typography";
+
+// Code that isn't truly approved yet
+// but needs to be implemented to try
+@import './release-candidate/index.scss';
```
stephenp: Import order needed to be adjusted for inheritance

src/styles/release-candidate/components/_c-button.scss
```diff
@@ -1,21 +1,23 @@
 .c-button {
```
stephenp: Removed a bit of the dry sass mixins and helpers until we build out the components more, it feels like overkill to have for just a few different variations.

src/styles/release-candidate/components/_c-button.scss
```diff
@@ -88,41 +82,57 @@
     }
 
     .loader {
+      position: absolute;
```
stephenp: Loader css moved inside of button `--loader` for scoping

Issue #22, opened 2018-06-19
Create footer component


bernabe9:

stephenp:
Merged in: https://github.com/yankaindustries/mc-components/pull/26

PR #26, opened 2018-06-19
Adding base footer component

stephenp:
First pass at the footer component.

Will need some help figuring out the best way to build this out, for now it's mainly static markup.  Brings in the styles from mc-fe-styles for the existing footer for now.

## Risks
None / Low

## Changes
Adds new footer component

## Issue
Footer component: https://github.com/yankaindustries/mc-components/issues/22


src/components/Footer/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+import logo from '../../assets/logo.svg'
+import sslBadge from '../../assets/ssl-badge.svg'
+
+export default class Footer extends PureComponent {
+  infoBlock = React.createRef()
+
+  render () {
+    return (
+      <div id="mc-footer">
```
bernabe9: should we use `id`s for the html elements? I think is a good practice to use classes instead of `id`s.. But I'm not sure if there is a good reason to use `id`s here

src/components/Footer/index.js
```diff
@@ -0,0 +1,63 @@
+import React, { PureComponent } from 'react'
+import cn from 'classnames'
+
+import logo from '../../assets/logo.svg'
+import sslBadge from '../../assets/ssl-badge.svg'
+
+export default class Footer extends PureComponent {
+  infoBlock = React.createRef()
```
bernabe9: This variable is not used. please remove it

src/components/Footer/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+  render () {
+    return (
+      <div id="mc-footer">
+        <div class="mc-container mc-contain-inner">
+          <img src={logo} alt='Logo wordmark' id="logo-footer" />
+
+          <div class="nav-footer">
+            <ul class="nav-footer-items">
+              <li class="nav-footer-item">
```
bernabe9: It would be great to move this to another component called `NavFooterItem`

Something like:

```javascript
const NavFooterItem = ({ label, url, external = false }) => (
  <li class="nav-footer-item">
    <a
       href={url}
       target={external ? '_blank' : '_self'}
     >
       {label}
    </a>
  </li>
)

export default NavFooterItem
```

src/components/Footer/index.js
stephenp: I implemented the header and footers with IDs originally because there was only ever supposed to be one and we'd have issues if there were more than one on the page.  If we think we could possibly use two footers in the future, I can update these to classes no problem, but I think for something as unique as a header and footer, specifically "the" header and footer, IDs are correct semantically

src/components/Footer/index.js
stephenp: I'm open to pairing on this to implement, I wasn't sure how much modularity one should use in a react component - i'm still new to this!  If you're available to pair, please let me know, but for now is this ok to implement as-is?

src/components/Footer/index.js
stephenp: Removed with next commit...

stephenp:
Lint errors fixed (I think) and comments added!

src/components/Footer/index.js
bernabe9: Makes sense

Issue #36, opened 2018-06-22
prettier-eslint isn't formatting correctly


bernabe9:
It's changing the JSX alignment.

Example:
1- change to a new branch: `git checkout -b test`
2- modify a line in the file `/src/components/Button/index.stories.js`
3- run git add
4- commit the changes: `git commit -m "test"`

Result:

```javascript
storiesOf('Button', module)
  .add(
    'primar',
    withInfo({ ...infoProps })(() => (
      <Button onClick={action('clicked')}
primary
      >
        Button
      </Button>
    )),
  )
  .add(
    'secondary',
    withInfo({ ...infoProps })(() => (
      <Button onClick={action('clicked')}
secondary
      >
        Button
      </Button>
    )),
  )
  .add(
    'tertiary',
    withInfo({ ...infoProps })(() => (
      <Button onClick={action('clicked')}
tertiary
      >
        Button
      </Button>
    )),
  )
  .add(
    'text button',
    withInfo({ ...infoProps })(() => (
      <Button onClick={action('clicked')}
text
      >
        Button
      </Button>
    )),
  )
  .add(
    'with icon',
    withInfo({ ...infoProps })(() => (
      <Button className='c-button--facebook'
Icon={<FacebookIcon />}
      >
        CONNECTED
      </Button>
    )),
  )
  .add(
    'loading',
    withInfo({ ...infoProps })(() => (
      <Button onClick={action('clicked')}
primary
loading
      >
        Button
      </Button>
    )),
  )
```

It seems to be a problem with the `@masterclass/eslint-config`, because when I remove that extension from `.eslint` file it works.

bernabe9:
The same issue happens with other js files

stephenp:
We should also exclude SVGs from being linted!

https://github.com/yankaindustries/mc-components/pull/35/files#diff-cdd49b00bce39a17f6bdb1c4404dd279

stephenp:
Moved to https://github.com/yankaindustries/mc-components/issues/44

PR #37, opened 2018-06-22
Temporal disable prettier-eslint and lint --fix

bernabe9:
### What
Temporal disable prettier-eslint and eslint --fix, we should enable it again once we tackle this issue #36 
### Why:
`prettier-eslint` is uglifying the code

## Risks
No

## Changes
No visual changes

## Issue
#36 and #37

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,8 +63,7 @@
   },
   "lint-staged": {
     "src/**/*.js": [
-      "eslint --fix",
-      "git add"
+      "yarn lint:js"
     ],
     "src/**/*.scss": [
       "stylelint --fix",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -63,9 +63,6 @@
   },
   "lint-staged": {
     "src/**/*.js": [
-      "prettier-eslint --write --log-level error",
-      "eslint --fix",
```
stephenp: Shouldn't it still run `eslint` just not `--fix`?

package.json
```diff
@@ -63,9 +63,6 @@
...
-      "git add"
```
stephenp: And then we would keep this as well

package.json
bernabe9: You're right

a-axton:
What do you mean by uglifying? Prettier should make things prettier 💅

PR #43, opened 2018-06-22
VideoPlayer component

a-axton:
Added a VideoPlayer component for brightcove video players. It's mostly a proof of concept until we try to use it in a feature. It may need some changes to make it more redux friendly, eg allowing props to be updated after creation and adding prop for pausing/playing the video.

## Risks
None

## Changes
No changes to existing components

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -1,0 +1,203 @@
+import React from 'react'
+import cn from 'classnames'
+import PropTypes from 'prop-types'
+
+export default class VideoPlayer extends React.PureComponent {
+  static defaultProps = {
+    playerId: 'rkcQq7gAe',
+    videoId: '5450137526001',
+    theme: 'default',
+    isLooped: false,
+    isMuted: false,
+    hasAutoplay: true,
+    hasControls: true,
+    hasBreakpoints: false,
+  }
+
+  static propTypes = {
+    playerId: PropTypes.string.isRequired,
+    videoId: PropTypes.string.isRequired,
+    /** Player theme styles, must have correlating styles to go with it */
+    theme: PropTypes.oneOf(['default', 'chapter']),
+    /** Pass in a react component to be shown at the end of the video. */
+    endscreenComponent: PropTypes.element,
+    /**
+     * Keeps video constrained to standard sizes, eg 1080x720.
+     * Used to keep videos looking crisp.
+     * Scales responsively.
+     * */
+    hasBreakpoints: PropTypes.bool,
+    hasControls: PropTypes.bool,
+    isMuted: PropTypes.bool,
+    isLooped: PropTypes.bool,
+    hasAutoplay: PropTypes.bool,
+    /** Player has been initialized */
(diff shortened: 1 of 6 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  renderEndScreen = () => (
+      <div className='bc-player-endscreen'>
+        <div
+          className='bc-player-endscreen__replay'
+          onClick={this.handleReplayClick}
+        >
+          <svg
```
bernabe9: Can you move this svg to a component inside the folder `/src/components/icons`?

src/components/VideoPlayer/index.js
```diff
@@ -0,0 +1,203 @@
...
+        </div>
+      </div>
+  )
+
+  render () {
+    return (
+      <div
+        className={cn('bc-player', {
+          'bc-player--endscreen-open': this.state.endscreenOpen,
+          'bc-player--has-breakpoints': this.props.hasBreakpoints,
```
bernabe9: use object destructuring for all the props inside props to avoid repeting `this.props`

for example:

```javascript
render () {
  const { hasBreakpoints, endscreenComponent, .... } this.props
}
```

src/components/VideoPlayer/index.js
```diff
@@ -0,0 +1,203 @@
...
+        })}
+      >
+        {this.props.endscreenComponent && this.renderEndScreen()}
+        <div className='bc-player__wrapper'>
+          <video
+            data-application-id
+            ref={(ref) => {
```
bernabe9: use the React 16 approach to create refs: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

src/components/VideoPlayer/index.js
```diff
@@ -0,0 +1,203 @@
...
+            }}
+            className={cn(
+              'video-js',
+              'bc-player__video',
+              `bc-player__video--${this.props.theme}`,
+            )}
+            data-embed='default'
+            data-video-id={this.props.videoId}
+            data-player-id={this.props.playerId}
+            data-account='5344802162001'
```
bernabe9: You also used this value in this [line](https://github.com/yankaindustries/mc-components/pull/43/files#diff-c00393e43603fb0a9c25af3001fe7719R8), what do you think about move it to a constant?

src/components/VideoPlayer/index.js
```diff
@@ -0,0 +1,203 @@
...
+      })
+      this.video = window.videojs(this.playerRef)
+      this.video.ready(this.handlePlayerReady)
+    }
+  }
+
+  handlePlayerReady = () => {
+    if (this.props.onPlay) {
+      this.video.on('play', () => {
+        this.props.onPlay(this.video)
```
bernabe9: use object destructuring for the props inside `this.props`

src/components/VideoPlayer/index.js
a-axton: Those are actually different numbers, they looks super similar though

PR #47, opened 2018-06-27
Add left property to header component

nicastelo:
Add `left` property to header component so that the header can show content at it's left

## Risks
None


src/components/HeaderLoggedIn/index.js
```diff
@@ -46,6 +52,28 @@ export default class HeaderLoggedIn extends PureComponent {
     this.setState({ isMobileMenuOpened: !isMobileMenuOpened })
   }
 
+  renderLeft = () => {
+    const { left } = this.props
```
bernabe9: move `type`, `action` and `label` to this object destructuring to avoid repeating `left`:

```javascript
const { type, action, label } = this.props.left
```

src/components/HeaderLoggedIn/index.js
```diff
@@ -46,6 +52,28 @@ export default class HeaderLoggedIn extends PureComponent {
...
+
+    if (left.type === 'back') {
+      return (
+        <button
+          className='header__left-button'
+          onClick={typeof left.action === 'function' && left.action}
```
bernabe9: You are already validating the `left.action` type with the proptypes.

Try with: `onClick={Boolean(action) && action }`

Anyway, this is just a suggestion. No need to change it

src/components/HeaderLoggedIn/index.js
```diff
@@ -73,8 +101,9 @@ export default class HeaderLoggedIn extends PureComponent {
           onClick={this.toggleMobileMenu}
         />
         <nav className='header__nav'>
-          <div className='header__section'>{/* TODO: breadcrumbs */}</div>
-
+        <div className='header__section'>
+          { this.props.left && this.renderLeft() }
```
bernabe9: Move `left` to the above object destructuring

Issue #46, opened 2018-06-27
Build grid component example


stephenp:
Build out an example component showing our grid system similar to [bootstrap's documentation](https://getbootstrap.com/docs/4.0/layout/grid/).  The designers would like to see how it works in practice with their values before confirming final values.

stephenp:
I'm not sure the best way to do this - build a custom component just for showing a grid layout seems like overkill and unnecessary - thoughts?

bernabe9:
I think that it's an excellent idea to have an example of how to use the grid.

Maybe the best solution is to have another storybook for non-components styles. But for now, we can add it as you said using a custom component with a story to show the grid.

jstnjns:
I know I'm a little late to the game, but per the convo at #16 I don't think we need a React component, but to just have the story include static HTML.

stephenp:
Closing this issue since the component has been created, but we can continue this discussion in the following issue to flesh out the static markup section in mc-components:

https://github.com/yankaindustries/mc-components/issues/16

PR #49, opened 2018-06-28
Adding grid example component

stephenp:
Add a sample component to demonstrate the grid system with a few examples.

## Risks
No

## Changes
No visual changes to existing components

## Issue
Addresses https://github.com/yankaindustries/mc-components/issues/46

### DIFF
--- a/src/styles/components/_grid-example.scss
+++ b/src/styles/components/_grid-example.scss
@@ -1,0 +1,8 @@
+.grid-example {
+  &-block {
+    background: $mc-color-primary;
+    width: 100%;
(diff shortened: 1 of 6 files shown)


src/styles/components/_grid-example.scss
```diff
@@ -0,0 +1,8 @@
+.grid-example {
```
bernabe9: Given that this classname only used to demonstrate the grid system and we don't want to include it in the compiled css, what do you think about moving this to inline styles inside the `GridExample` component?

Issue #13, opened 2018-06-17
Add Chapter Video Tile Component


jesseocon:
## Description:
The chapter video displays the thumbnail of the chapter video and after .5s of hover is scales and starts playing the video. The video has sound and no controls. clicking on the video goes to the the chapter of the video playing.

### Requirements
* Should take a thumbnail image
* Should take a link address (for linking to the chapter)
* Should take a video hash
* Should take a video vendor type: (wistia, brightcove)

### Open questions
* Should the scaling be included in the component or should that be a something added via render props. I think it is currently the latter.

### Design file
https://www.figma.com/file/KQr5QmLmvKzpOaxp8oEBoXE6/Logged-In-HP

### Current implementation
The current implementation can be seen by: 
1. Clone: https://github.com/bernabe9/mc-assignment
2. If you don't have access slack Bernabé
3. yarn start
4. Hover any tile in **my classes**.

![2018-06-17 07 58 17](https://user-images.githubusercontent.com/1948630/41509143-5478f622-7204-11e8-8c17-eb224655f40e.gif)

jstnjns:
This is now possible with Tile, TileImage, TileCaption, HoverHandler, and AnimationHandler.  See Storybook for details

PR #50, opened 2018-06-28
Added VideoTile component and more components that we need to render …

stabarez:
<General description>

Added VideoTile component and more components that we need to render it.

## Risks
No risks.

## Changes

- Added ScaleContainer.
- Added BrightcoveVideo component.
- Added VideoProgress component
- Added TileCaption component
- Added AudioButton component.
- Added VideoTile component.
- Added instructions to the README.

## Issue
No Issue. Feature associated: https://github.com/yankaindustries/mc-components/issues/13


src/components/ScaleContainer/index.js
```diff
@@ -0,0 +1,119 @@
...
+
+    return (
+      <div
+        className={`${className} ${hoverClass} ${zoomClass}`}
+        style={style}
+        ref={this.containerRef}
+        onMouseEnter={this.onMouseEnter}
+        onMouseLeave={this.onMouseLeave}
+      >
+        {children(Boolean(zoomClass), reverseScale)}
```
jstnjns: Instead of passing these in as a list of arguments, I think it might be better as an object.  That way, if we ever add to this list in the future, the order is irrelevant and we can use destructuring to get what we need without having to pull everything before it.

src/components/TileCaption/index.js
```diff
@@ -0,0 +1,50 @@
...
+        className={`
+          tile-caption
+          tile-caption--${position}
+          ${animated ? 'tile-caption--animated' : ''}
+        `}
+        style={style}
+      >
+        <div className='tile-caption__container'>
+          {comingSoon &&
+          <p className='tile-caption__coming-soon'>Coming soon</p>
```
jstnjns: I feel like we should also make this a prop so that we are not painting ourselves into a corner of only being able to have "Coming soon" here.  `suptitle` maybe?

jstnjns:
I forgot to mention, but check out the prefix for stories (https://github.com/yankaindustries/mc-components/blob/develop/src/components/Modal/index.stories.js#L40) for organizing the navigation.

stabarez:
@jstnjns I will review the position/size flickering. I think that it's related to how I'm rendering the component on storybook, becase I integrated these components on this branch on Masterclass: feature/mc-components-integration-test (far away from develop by now) and everything was fine.

src/components/ScaleContainer/index.js
stabarez: Make sense.

src/components/TileCaption/index.js
stabarez: I agree.

jstnjns:
I think this is solved with our new Tile components.  Let me know if you think the new implementation is missing something.

PR #52, opened 2018-06-29
Tests: Set up storyshots

### LINKED ISSUE #4: Add test for all the existing components
mc-components now is official and we need to have test for our components.

Snapshot testing should be enough for components without logic, otherwise we need to add unit tests

bernabe9:
Add snapshots for all the stories using Storyshots.

## Risks
Noe

## Changes
- Set up Jest
- Set up Enzyme
- Set up Storyshots

## Issue
Resolves #4

### DIFF
--- a/src/components/Carousel/index.js
+++ b/src/components/Carousel/index.js
@@ -17,7 +17,7 @@ const defaultSettings = {
 
 export default class Carousel extends Component {
   static propTypes = {
-    data: PropTypes.arrayOf(PropTypes.shape({})),
+    slidesDataSource: PropTypes.arrayOf(PropTypes.shape({})),
     children: PropTypes.element,
     onSlideClick: PropTypes.func,
     settings: PropTypes.shape({}),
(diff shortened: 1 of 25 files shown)


src/components/Carousel/index.js
```diff
@@ -17,7 +17,7 @@ const defaultSettings = {
 
 export default class Carousel extends Component {
   static propTypes = {
-    slidesDataSource: PropTypes.arrayOf(PropTypes.shape({})),
+    data: PropTypes.arrayOf(PropTypes.shape({})),
```
nicastelo: Is this intencional or was overwritten by the merge?

bernabe9: Intentional, to fix the carousel.
Isn't working now, you're using the prop `data` in the story:

screen shot 2018-07-02 at 10 52 34 am

But instead of changing the prop name in the component we can change the story to inject the prop `slidesDataSource`. What do you prefer?

nicastelo: Oops. I think the change in the story will do better

bernabe9: Fixed

PR #77, opened 2018-07-10
Add mailcheck option to input

missx:
On masterclass I worked on a story to add mailcheck to an email input. This is the url: https://github.com/yankaindustries/masterclass/pull/6172

I wanted to add it as an option for the input because in the other PR I had to create another input to develop the story and remove the one from mc-components.

## Risks

None

## Changes

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "bootstrap": "^4.1.1",
     "classnames": "^2.2.5",
+    "react-mailcheck": "^1.2.0",
     "react-slick": "^0.23.1",
     "slick-carousel": "^1.8.1"
   },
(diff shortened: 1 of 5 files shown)


bernabe9:
I'm not sure about this...

Why? We are adding dependencies (`react-mailcheck`)

I like the idea of creating a story showing how mc-component's Input works with `react-mailcheck` but not to include that inside the library.

Check out how MaterialUI does to show examples with other libraries integrations: https://material-ui.com/demos/text-fields/#formatted-inputs

missx:
@bernabe9 You mean to do that in the masterclass PR?

jstnjns:
I agree with @bernabe9, I don't think it should be included in the component library to include things like validation.  I would go a step further and say that the implementation in `masterclass` should make use of `redux-form`'s validation mechanism (assuming that `redux-form` is being used).

missx:
Agreed. I managed to use the mailchecker library and not the react-mailchecker in masterclass, so I'm going to close this PR and open another one but just to add the warning prop to the Input component.

PR #80, opened 2018-07-11
Add warning to input component

missx:
Add a warning prop to the input component.

## Risks

None

## Changes

image

### DIFF
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -47,6 +47,7 @@ export default class Input extends PureComponent {
       { 'input-field__input--with-label': showLabel },
       { 'input-field__input--full-width': fullWidth },
       { 'input-field__input--error': error },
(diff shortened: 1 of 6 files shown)


src/components/Input/index.js
```diff
@@ -60,6 +62,9 @@ export default class Input extends PureComponent {
         {error &&
           <span className='input-field__error'>{error}</span>
         }
+        {warning &&
+          <span className='input-field__error'>{warning}</span>
```
bernabe9: maybe change the class name to `input-field__warning`?

bernabe9: So the developers are able to override the styles outside the library with different classes

missx: You're right!

bernabe9:
@missx update your tests(`yarn test -u`) and push again to pass CircleCI

PR #87, opened 2018-07-12
FormatClassTitle helper & export helpers

nicastelo:
Added a helper built by @stabarez (FormatClassTitle) that was in the packs of the main repo. Moved to here in order to reuse

Also, exporting helpers in order to use them from outside


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mc-components",
   "version": "0.2.0",
-  "main": "./dist/components/index.js",
+  "main": "./dist/index.js",
```
bernabe9: The file `./dist/index.js` doesn't exist

nicastelo: Good catch, I'll upload the dist files too?

PR #89, opened 2018-07-13
Add Foundation component for typography

stephenp:
- Update PR template since we removed all merge types except squashing
- Updated grid naming to reflect "foundation" in code
- Added type component to foundation grouping
- Remove storybook style overrides

## Risks
Risk around typography if these components are being used elsewhere - there was a root font size declaration to override to 1.6rem (for easy font scaling down to mobile).  Since the style guide doesn't showing consistently scaling fonts, I don't think it makes sense to use rems since all fonts will scale at different rates.

## Changes
Just adding a new component!

## Issue
N/A


.storybook/styles.scss
```diff
@@ -1,20 +1,6 @@
...
-h2,
-h3,
-h4,
-h5,
-h6 {
-  color: #000;
-}
+// Overrides masterclass default bg color
+// so that storybook can change bg
+body { background: transparent !important; }
```
jstnjns: kill this too, we should move to not have a changeable background as it adds complexity without much benefit

src/components/FoundationGrid/index.stories.js
```diff
@@ -2,12 +2,12 @@ import React from 'react'
...
-import GridExample from '../GridExample'
+import FoundationGrid from '../FoundationGrid'
 
-storiesOf('foundation|GridExample', module)
+storiesOf('foundation|Grid', module)
   .add(
     'default',
     withInfo()(() =>
-      <GridExample />,
+      <FoundationGrid />,
```
jstnjns: Check my PR, we can actually just use HTML here.  Having a component for this hides the markup in the documenation.

src/components/FoundationType/index.stories.js
```diff
@@ -0,0 +1,13 @@
...
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+
+import FoundationType from '../FoundationType'
+
+storiesOf('foundation|Type', module)
+  .add(
+    'default',
+    withInfo()(() =>
+      <FoundationType />,
```
jstnjns: Same note I have with the Grid, just put your markup right here.

stephenp:
@jstnjns Your requested changes have been implemented - I kept getting weird linting errors about needing a trailing comma, so that's why there's that weirdness going on in index.stories.js for foundationGrid and foundationType

jstnjns:
Just need to update your snapshots and you're good to go!

PR #90, opened 2018-07-13
Navigational structure

jstnjns:
Organizational changes, grouping components that are similar so that our navigation doesn't get astronomically long and difficult to skim through.

Automated documentation is convenient, but does not always produce the best results.  With the creation and utilization of a few helper methods, we can create more helpful, visually appealing documentation.  This PR introduces two components: `DocSection` and `PropExample`.

#### `<DocSection>`

This is pretty empty, it just adds a title to the section.  But this can get expanded on later if we want to make them collapsible or whatnot.

#### `<PropExample>`

The meat and potatoes of how we can make our documentation more interesting.  This currently takes a couple props:

- `name` prop name that is being featured
- `type` prop type
- `children` code example (gets rendered, as well as shown as code via toggle)

#### Result
screen shot 2018-07-12 at 3 56 52 pm

#### Implementation
screen shot 2018-07-12 at 4 13 07 pm

## Risks
Temporary fragmentation of the current method of displaying components (until other sections are updated)
Less automated process for generating documentation (with the promise of offering higher quality documentation)

## Changes
Starting work to remove `withInfo` wrapper for stories
Updates Button docs as example
Organizes components into groups that make sense and are easy to navigate through


jstnjns:
@bernabe9 I figure for those we’d just wrap the example with a white background. Having to switch manually isn’t a great experience for the user, so we can just add some extra UI to the docs as needed. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #108, opened 2018-07-23
Removing unused legacy icons

stephenp:
## Overview
I manually copied over the icons we used to use in mc-fe-styles, but they should be removed from the styles folder until needed, at which point they should live under `assets/icons`.

## Risks
Did a "find" in mc-components for each of the images, no matches found, they still exist in mc-fe-styles so there should be no breaking for masterclass repo either.

## Changes
Deletes all icon files in styles folder

## Issue
N/A


stephenp:
Checked main mc repo to see if these images were being referenced as well: 
image

Issue #76, opened 2018-07-10
HoverHandler


jstnjns:
Similar to the implementation of ScaleContainer but a bit more abstract, this component should have render props and classes for hover/nonhover states.  This allows us to change the contents of a container based on hover, as well as add a class for hover affects (like scaling).

PR #111, opened 2018-07-23
Refining hover and zoom functionality

jstnjns:
## Overview
Instead of having each component maintain their own code for handling animation on hover, I've built a couple utility components that should simplify this interaction.

## Risks
Currently none.  But if a designer wants some one-off animation, we might have to wrangle them in to use our predefined animations (or add another).

## Changes
zoom

## Issue
#76


src/styles/components/_tile.scss
```diff
@@ -45,3 +45,17 @@
 .tile--519x187 {
   @include aspect-ratio(519, 187);
 }
+
+
+.tile__caption {
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  margin: 1em;
```
stephenp: Should we use `em` or `px`? I've noticed that most browsers still have issues with animations being blurry if they land on non-whole pixel values (which can happen with `em`).

src/styles/animation/_animation.scss
```diff
@@ -0,0 +1,40 @@
+.animation {
```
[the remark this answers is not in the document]
jstnjns: Yea, that's a good idea.  Should all styles be prefixed then?  (or are they and I'm just not paying attention?)

src/styles/components/_tile.scss
jstnjns: Sounds good

src/styles/animation/_animation.scss
```diff
@@ -0,0 +1,40 @@
...
+
+    &.mc-animation--animating > * {
+      transform: scale(1.1);
+    }
+  }
+
+  &.mc-animation--ken-burns {
+    overflow: hidden;
+
+    > * {
```
stephenp: This might cause issues in the future if we have multiple direct children of the `mc-animation` parent (it will animate each child individually instead of the parent)

src/styles/animation/_animation.scss
```diff
@@ -0,0 +1,40 @@
+.mc-animation {
+  position: relative;
+
+  &.mc-animation--animating {
```
stephenp: BEM all of these, this should work as just `--animating`

src/styles/animation/_animation.scss
jstnjns: For reference for everybody else, Pontes and I cleared this up, that the component only accepts a single child, and that we could potentially add a wrapper later if we find it's necessary.

Issue #112, opened 2018-07-23
Replace old buttons in masterclass repo


stephenp:
We need to continue to build out the button component's functionality to support the various use cases in the masterclass repo.  Steps to complete this task:

- Any existing hard coded `mc-btn` or `mc-button` should be replaced with the new mc-component button.  
- mc-button style removed from mc-fe-styles
- mc-btn style removed from masterclass repo
- class name on mc-component button changed from `c-button` to `mc-button`

stephenp:
As pages get replaced with react components, that will be the time to update this.  Until then, it doesn't make sense to bring in all of mc-components on a page to replace a single button.

PR #113, opened 2018-07-23
Tech debt: button cleanup

stephenp:
## Overview
There are a number of different legacy button types now in the main masterclass repo. We have `mc-btn`, `mc-button`, and now `c-button`, all of which have different variants that are slightly different.

This PR is the first step towards finally cleaning that up.  This pr adds a temporary class that mirrors `c-button` called `mc-c-btn`.  This is a unique enough name to do a find and replace after cleanup is complete.  My plan is to:

- Migrate all existing `mc-btn` and `mc-button` classes to those supported by `c-button` by using the new temporary `mc-c-btn` class. 
- Once **all** buttons are consolidated in the masterclass repo, I'll find and replace all instances of `mc-c-btn` to `mc-btn` and update the component class names as well.
- From this point on, `mc-btn` will be the only button type.

This PR also adds linting to the precommit step so we won't forget to check CSS before committing code!

## Risks
Low risk now, as this is adding a new class name.  High risk later when we remove all existing class names.

## Changes
No visual changes

## Issue
Starting to attack: https://github.com/yankaindustries/mc-components/issues/112

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
       "yarn lint:js"
     ],
     "src/**/*.scss": [
+      "yarn lint:style",
       "stylelint --fix",
       "git add"
     ]
--- a/src/styles/components/buttons/_button--types.scss
+++ b/src/styles/components/buttons/_button--types.scss
@@ -1,4 +1,5 @@
-.c-button {
+.c-button,
+.mc-c-btn {
   &--small {
     padding: 12px 16px;
   }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -85,6 +85,7 @@
       "yarn lint:js"
     ],
     "src/**/*.scss": [
+      "yarn lint:style",
```
jstnjns: Doesn't `lint:style` run `stylelint`?

stephenp: https://github.com/yankaindustries/mc-components/blob/develop/package.json#L31

It does indeed!

Issue #66, opened 2018-07-08
ImageTile - backgroundGradient shouldn't be a prop


jstnjns:
If we think about what the background gradient is, it's really to allow content to hover over the image with better visibility / higher contrast.  Since we should be able to use the same mechanism for adding the background gradient across all components that utilize ImageTile, this could become problematic when we start to introduce more complexity (e.g. when the hover state might not have gradient, we then have to introduce props to child components letting them know if the component is being hovered on).

I would the gradient being, to start, pure CSS / a `className` that we can add to a wrapper.  We should also consider the idea that inline CSS, especially if not dynamically fed by values, should not exist.

jstnjns:
This was fixed with #117, as `ImageTile` is scheduled to be deprecated and Tile, TileImage, and TileOverlay are the preferred methods for building the same functionality.

PR #117, opened 2018-07-26
Super flexible Tile components

jstnjns:
## Overview
There's a fine balance between building a unique tile for every need/context, and having generalized tiles that have wayyyy too many props to get them to look like they need to in unique cases.

I submit to you, the start of a collection of composable tile elements that allow you to choose from a collection of common parts that make up tiles, but don't limit you to what is already built into the component.  For instance, if you are building a tile and the designer has introduced a new block of text that current tiles don't support you'd have two options: build the tile from scratch (even though maybe 80% of the pattern exists in existing components) or hack away at an existing component with conditional props to get your bespoke solution.

screen shot 2018-07-25 at 11 50 43 pm
screen shot 2018-07-25 at 11 50 47 pm
screen shot 2018-07-25 at 11 50 33 pm
screen shot 2018-07-25 at 11 50 38 pm

## Risks
ImageTile is currently used.  However, I didn't touch ImageTile, but instead created one that sounds similar (TileImage).  We should eventually phase out ImageTile because a combination of Tile/TileImage/TileOverlay/TileCaption does the exact same thing but with more flexibility.

## Changes
No changes to existing components, but a change to the way we build tiles in the future.

## Issue
This PR makes #66 irrelevant because we'd just deprecate ImageTile altogether.


src/components/Tile/index.stories.js
```diff
@@ -0,0 +1,113 @@
...
+import TileCaption from '../TileCaption'
+import TileCheck from '../TileCheck'
+import AnimationHandler from '../AnimationHandler'
+import HoverHandler from '../HoverHandler'
+
+import shondaRhimesThumbnail from '../../assets/shonda-rhimes-video.png'
+
+
+storiesOf('components|Tiles', module)
+  .add('Summary', () => (
```
jstnjns: There's a lot in this PR, but this story should hopefully explain most of it.

jstnjns:
It wouldn't let me post this gif last night.  But here's an example of how you can stack the parts to make a whole.

tile

jstnjns:
Here's some direction I already see:

TileVideo gets added
TileOverlay supports more types (fill, gradient-top, etc)
TileCaption takes typography options (title-level="h2") and positioning props (left-bottom, center-bottom, center)

stephenp:
I don't feel qualified to weigh in on a change this large, but the class name implementations for hover and animation events looks good to me.

Issue #14, opened 2018-06-17
Add Trailer Video Tile Component


jesseocon:
## Description
In the explore classes area of the logged in homepage when the user hovers over a tile for .5second the video tile scales, a button becomes visible and the trailer video starts to play. 

### Requirements
* Should take a thumb url for the poster image
* Should take a link address for where the tile should link
* Should take a callback for click handler

### Current implementation
The current implementation can be seen by: 
1. Clone: https://github.com/bernabe9/mc-assignment
2. If you don't have access slack Bernabé
3. yarn start
4. Hover any tile in **explore classes**.
![2018-06-17 08 10 27](https://user-images.githubusercontent.com/1948630/41509310-a514e134-7206-11e8-94ee-e8e369f27f5c.gif)

jstnjns:
This is solved with new `Tile` implementation

PR #124, opened 2018-07-31
Adding TileVideo

jstnjns:
## Overview
Adding video as a tile layer so that we can support more tile types.  

tile

## Risks
None

## Changes
Adding another component, TileVideo.

## Issue
This supports #13 and #14


src/components/ClassCard/index.stories.js
```diff
@@ -1,54 +0,0 @@
-import React from 'react'
```
jstnjns: ClassCard will still be available for those that need it or are currently using it.  But so that we opt for context-less components, I am removing it from the documentation.

src/components/ImageTile/index.stories.js
```diff
@@ -1,42 +0,0 @@
-import React from 'react'
```
jstnjns: Same with ImageTile, let's opt for
```
<Tile>
  <TileImage imageUrl={...} />
</Tile>
```
instead.

Issue #133, opened 2018-08-06
Type styles are too specific when importing into another codebase


stephenp:
The typography styles target h1 - h6 and the paragraph tag, which leads to style conflicts when `mc-components` is imported into another codebase.  We should only use classes to specify heading and body styles.

PR #134, opened 2018-08-06
Updating typography to not target direct elements

stephenp:
## Overview
The typography styles target h1 - h6 and the paragraph tag, which leads to style conflicts when `mc-components` is imported into another codebase.  We should only use classes to specify heading and body styles.

## Risks
Low - I don't believe anywhere is relying on the styles in mc-components for type yet.

## Changes
No visual changes

## Issue
https://github.com/yankaindustries/mc-components/issues/133


jstnjns:
JK

PR #142, opened 2018-08-08
Updating documentation type to use new naming

jstnjns:
## Overview
Documentation headers were using old classes, this updates to use new.

## Risks
None

## Changes
n/a

## Issue
n/a


dist/utils/PropExample.js
```diff
@@ -52,7 +52,7 @@ var Definition = function (_PureComponent) {
           { className: 'row' },
```
jstnjns: We should consider splitting out utils into two types...ones that we export and ones that we don't.

Issue #165, opened 2018-08-13
Modal updates


bernabe9:
### Change modal's classname to prevent collisions
Currently we are using the classname `modal`. In MasterClass we are already using the classnames: `modal` and `mc-modal`

### Custom mount element
The default element is `modal-root`.

### Automatically create the mount element if it doesn't exist
Currently it's necessary to include the mount element in the html structure, otherwise the modal portal will fail.

jstnjns:
Appears to have been done

PR #169, opened 2018-08-13
Modal Updates

nicastelo:
## Risks
Medium

## Changes

- Custom mountEntry
- Custom close button position
- Change class to avoid style conflicts with other modals in masterclass

screen shot 2018-08-13 at 4 08 24 pm

## Issue
#165


src/components/Modal/index.js
```diff
@@ -62,32 +82,53 @@ export default class Modal extends Component {
...
       HeaderComponent,
+      closeButtonPosition,
+      mountEntry,
       ...props
     } = this.props
+    const cnRoot = 'mc-components-modal'
 
     const classNames = cn(
-      'modal__container',
+      `${cnRoot}__container`,
```
jstnjns: Correct me if I'm wrong, @stephenp , but this should just be the root block `"mc-components-modal"` (which would just be `cnRoot`.

src/components/Modal/index.js
```diff
@@ -62,32 +82,53 @@ export default class Modal extends Component {
...
 
     if (!isOpen) {
       return false
     }
 
     return (
-      <ModalPortal>
-        <Close className='modal__close-icon' onClick={onClose} />
+      <ModalPortal mountEntry={mountEntry}>
+        { /outside/.test(closeButtonPosition) &&
```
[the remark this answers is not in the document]
stephenp: Yep - with a root class name on the component, we should be able to control a single close button element's position

src/components/Modal/index.js
stephenp: yeah, I think the class name should be something like `mc-modal` at the root, and then `mc-modal__container` inside if necessary

Issue #170, opened 2018-08-13
Make ButtonLoader to accept the color as prop


nicastelo:
Make ButtonLoader to accept the color as prop so that we can change the color through react

e.g.

```
<ButtonLoader color='#000' />
```

PR #171, opened 2018-08-13
ButtonLoader color prop

nicastelo:
## Overview
Make ButtonLoader to accept the color as prop so that we can change the color through react

## Risks
None

## Issue
#170


stephenp:
Adding @jstnjns since I think he was working on something similar in a different component.

Issue #164, opened 2018-08-10
SVGs are not bundled with dist files


stephenp:
Need to figure out solution...

Issue #143, opened 2018-08-08
Split utils by whether we should export them or not


jstnjns:
Right now, because we put utils in dist we are packaging documentation helpers.  Prolly shouldn't.

PR #174, opened 2018-08-15
Simplify build process and clean up exports

jstnjns:
## Overview
The build process had a few too many steps, this cleans them up and exports proper files and minimized the number of files added to `dist` output.

## Risks
High.

This touches a lot of the files that are imported, and should be checked in the `masterclass` repo before importing.  This should make it apparent where we are improperly importing `mc-components` and its parts, and hopefully simplify/correct how it's being used.

I was able to `yarn link` after making these changes, and importing seemed to work as expected, but I suspect it should still be thoroughly tested.
 
## Changes
Removes SCSS from `dist` (these can be imported from `src` if necessary)
Removes separately transpiled components from `dist` (these can be imported from `src` if necessary)
Assures inline SVGs work properly
Moved Babel config to `.babelrc`

## Issue
#164 
#143


package.json
```diff
@@ -66,6 +73,12 @@
     "react-dom": "^16.4.0",
     "react-highlight": "^0.12.0",
     "rimraf": "^2.6.2",
+    "rollup": "^0.64.1",
```
jstnjns: Adding Rollup.js and plugins.  While some of the plugins mirror what we were doing with Webpack, it seems to be the standard for React/UI/Module libraries (see Material UI, Bootstrap, etc).

package.json
```diff
@@ -83,17 +96,6 @@
       "git add"
     ]
   },
-  "babel": {
```
jstnjns: Moved to `.babelrc`

rollup.config.js
```diff
@@ -0,0 +1,51 @@
+import path from 'path'
+import svgr from '@svgr/rollup'
+import commonjs from 'rollup-plugin-commonjs'
+import babel from 'rollup-plugin-babel'
+import { terser } from 'rollup-plugin-terser'
+
+import pkg from './package.json'
+
+
+const globals = {
```
jstnjns: Globals don't get bundled.  Keeps our size down and prevents possible conflicts with multiple libs getting loaded.

package.json
```diff
@@ -1,54 +1,61 @@
...
+    "build:css": "node-sass src/styles/index.scss | postcss > dist/styles.css",
+    "build:css-vendor": "node-sass src/styles/libraries/_libraries.scss | postcss > dist/vendor.css",
+    "build:styles": "yarn build:css && yarn build:css-vendor",
     "start": "start-storybook -p 9001 -c .storybook",
     "deploy-storybook": "storybook-to-ghpages",
     "prepublishOnly": "yarn build",
     "lint:js": "eslint src --color",
     "lint:style": "stylelint \"src/styles/**/*.scss\" || echo \"Style lint failed.\"",
     "lint": "yarn lint:js && yarn lint:style",
-    "precommit": "lint-staged",
```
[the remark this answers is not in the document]
stephenp: I'm not sure I agree with this one - I like having linting done on commit because otherwise you need to remember to run it manually before pushing a PR.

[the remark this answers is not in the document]
jstnjns: CircleCI will still lint for you, but if you quickly want to save your work the linter stops you from being able to commit and switch to something else.  This, IMO, is a blocker.

https://ourmasterclass.slack.com/archives/GBC5JV6HF/p1532498175000023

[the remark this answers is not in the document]
jstnjns: If we can figure out how to have the linter show you the errors, but not block you from committing, I'm fine with that.  But blocking is 👎

.storybook/config.js
```diff
@@ -20,7 +20,7 @@ setDefaults({
   inline: true,
 })
 
-const req = require.context('../src/components', true, /\.stories\.js$/)
+const req = require.context('../src', true, /\.stories\.js$/)
```
jstnjns: Moved `playground` and `foundation` outside of `components`

.gitattributes
```diff
@@ -0,0 +1 @@
+dist/* linguist-generated
```
jstnjns: This is to automatically collapse dist files in PRs

package.json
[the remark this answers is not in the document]
a-axton: you can use `--no-verify` to commit without linting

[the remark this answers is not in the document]
jstnjns: Good to know.

prettier.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
jstnjns: This is for SVGR

Issue #183, opened 2018-08-28
Body Styles


jstnjns:
We should attribute a base background as well as font color to the body.  This shouldn't really affect existing use cases, as they mostly set/override these values anyway.

<img width="1378" alt="screen shot 2018-08-28 at 10 03 16 am" src="https://user-images.githubusercontent.com/249444/44738668-74783f80-aaaa-11e8-96c9-68113ec99ab2.png">

All of these values should exist as variables.  My assumption is that we use our existing `$mc-background(-color)`, add `$mc-text-color` and `$mc-header-color`, and then maybe add `$mc-background-color-inverted` and `$mc-text-color-inverted` as well (for the white sections).

PR #187, opened 2018-08-29
Define color variables for body background and text colors

stephenp:
Defines color variables for background, text color, and sets them on body

## Overview
Adds new variables: 
```
$mc-color-background:        #090909 !default;
$mc-color-background-invert: #ebeeee !default;
$mc-color-text:              #272c33 !default;
$mc-color-text-invert:       #8ca1c1 !default;
```

And sets them by default on body - now when you import mc-components' styles into a page, you'll get the black background and white text by default.

## Risks
None - checked mc repo and mc-components for variable use and does not exist.

## Changes
N/A

## Issue
https://github.com/yankaindustries/mc-components/issues/183

### DIFF
--- a/src/styles/typography/base.scss
+++ b/src/styles/typography/base.scss
@@ -3,6 +3,7 @@ html {
 }
 
 body {
+  background: $mc-color-background;
   font-family: $mc-font-default;
   font-size: 15px;
   line-height: 1.5;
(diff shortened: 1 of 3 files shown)


src/styles/typography/base.scss
```diff
@@ -3,6 +3,7 @@ html {
 }
 
 body {
+  background: $mc-color-background;
```
jstnjns: no `color: $mc-color-text`?

stephenp: Shh....

stephenp: Check line 11 ;)

Issue #154, opened 2018-08-08
Better carousel abstraction


jstnjns:
Carousel is built well, but for a single use case.  I'd like to abstract it to three components.  One for heroes, one for tiles, and one that links them together.

```
<CarouselLinker>
  <CarouselHero>
    {courses.map(course => (
      <Hero background={course.image}>
        <h2>{course.title}</h2>
      </Hero>
    ))}
  </CourselHero>
  
  <CarouselTiles>
    {courses.map(course => (
      <Tile>
        <TileImage imageUrl={course.image} />
        <TileCaption title={course.title} />
      </Tile>
    ))}
  </CarouselTiles>
</CarouselLinker>
```

jstnjns:
🔥

Issue #114, opened 2018-07-24
Carousel settings should be separate props


jstnjns:
Current implementation is to pass a prop called settings, which is an object of the actual settings.  We should separate each of these settings into their own props (or, at least settings that we care to allow to be changed). This helps us validate the values with PropTypes.  It also means that if we change the implementation (say, use a different carousel mechanism) passing settings is hopefully less complicated.

PR #189, opened 2018-08-30
Improved Carousel

jstnjns:
## Overview
The existing carousel isn't so flexible.  Instead of spending a lot of effort to refactor it, I've created a new one with documentation on it's different modes.  The goal is to be able to replicate all the sliders we use throughout the site: home/hero, category slider, and anywhere else we might use it.

Additionally, I've added a component to connect carousels (ref/asForNav implementation of `react-slick`)

## Risks
Low.  The old carousel is still being used in `masterclass`, but all that needs to happen there is update the reference from

```
import { Carousel } from 'mc-components'
```

to

```
import { LegacyCarousel } from 'mc-components'
```

for a drop in replacement.  But it is advised to just update to the new carousel implementation.

## Changes
screen shot 2018-08-30 at 1 34 35 am
screen shot 2018-08-30 at 1 34 46 am

## Issue
#154
#114


src/components/Button/index.js
```diff
@@ -9,6 +9,7 @@ const Button = ({
   children,
   className,
   onClick,
+  primary,
```
bernabe9: why did you add this prop again?

jstnjns: ah, good catch.  to get the errors to stop.  it was meant to be temporary and never merged in.  i will fix this now

PR #194, opened 2018-08-31
Enhance video player end screen

bernabe9:
## Overview
We were adding a Replay Icon in the end screen component instead of allowing the user to inject a completely custom end screen component.

Now the prop `endscreenComponent` receives a component that will have a function `onReplay`.

## Risks
None. We are not using this the endscreenComponent on masterclass yet.

## Changes
No visual changes

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -243,7 +243,7 @@ export default class VideoPlayer extends PureComponent {
           >
             {renderChildren(
               endscreenComponent,
-              { onReplay: this.handleReplayClick })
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -145,7 +145,12 @@ export default class VideoPlayer extends PureComponent {
       this.video.play()
     } else if (endscreenComponent) {
       this.setState({ endscreenOpen: true })
-    } else if (pausescreenComponent) {
+      // eslint-disable-next-line
+      if (this.video.isFullscreen_) {
```
jstnjns: is there any downside to executing `closeFullscreen` if the video isn't in fullscreen?

Issue #178, opened 2018-08-24
Adjust typography based on screen size


jstnjns:
Font sizes should change based on what size display you're viewing from.  See components figma for more information: https://www.figma.com/file/X4sR3879DmwTYCRG72WKquAP/Components

PR #196, opened 2018-08-31
Adding responsive typography per design comp

stephenp:
## Overview
Resize text elements based on viewport width.  These values will most likely change, but add in support to easily update as the design comp becomes more finalized.

## Risks
Low - this will most likely visually affect our components as we tweak the font sizes with design.  We are ok with visual changes due to these updates.

## Changes
- Default body font got a slight size bump
- Text default is from 0 - "small" (576px)
- Next breakpoint is from "small" (576) - "large" (992px)
- Last breakpoint (biggest text) is from "large" (992px) and up.

## Issue
https://github.com/yankaindustries/mc-components/issues/178

### DIFF
--- a/src/styles/typography/base.scss
+++ b/src/styles/typography/base.scss
@@ -2,10 +2,11 @@ html {
   font-size: 62.5%;
 }
 
+// Referenced as "paragraph 2" in comps
 body {
   background: $mc-color-background;
   font-family: $mc-font-default;
-  font-size: 15px;
+  font-size: 12px;
   line-height: 1.5;
   font-weight: 400;
   color: $mc-color-text;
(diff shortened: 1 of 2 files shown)


src/styles/typography/base.scss
```diff
@@ -2,10 +2,11 @@ html {
   font-size: 62.5%;
 }
 
+// Referenced as "paragraph 2" in comps
 body {
   background: $mc-color-background;
   font-family: $mc-font-default;
-  font-size: 15px;
+  font-size: 12px;
```
stephenp: New default for desktop is 16px, but on mobile it's 12px

PR #206, opened 2018-09-13
Updating breakpoints

stephenp:
## Overview
There was a misalignment of expectations between design and development on how the grid works.  We spoke with Sean and got clarification on how design would like to have the grid implemented and these code changes reflect those discussions.

## THIS IS A POTENTIALLY BREAKING CHANGE
Any components using the existing grid might have edge cases while hitting breakpoints that could visually break them.  @jstnjns and I will be going through existing components to verify there are no breaking changes as well as the masterclass repo before versioning mc-components.  We would like to get this change in ASAP so any developers building new components will be using the grid now vs later.  There should be a separate PR in the next few days to adjust any existing components to work with the new grid.

## Things to check
Per Jesse, these items in the main masterclass repo may be affected by this change:
- Playlists and LIHP 
- 7 day lesson plan
- onboarding (red carpet)

## Risks
Medium - High, lots of potential issues with resizing components with the new grid.  The max widths and breakpoints that we were using before don't necessarily apply.

## Changes
Changes can be viewed on the grid storybook example.

When merged into masterclass, we should reference [[WI-21]](https://masterclass-dev.atlassian.net/browse/WI-21)
## Issue
N/A

### DIFF
--- a/.storybook/_upside-down.scss
+++ b/.storybook/_upside-down.scss
@@ -204,6 +204,41 @@ a {
   }
 }
 
+// Inline variables magically, thanks @jones for
+// the idea. You're so smart.
+ .mc-gutter-value {
+  &:before {
+    content: "#{$grid-gutter-width}";
+  }
+
+  &--xl:before {
+    content: "#{$grid-gutter-width-xl}";
+  }
+ }
+
+.mc-bp-value {
+  &--xs:before {
+    content: "#{$mc-bp-xs}";
+  }
+
+  &--sm:before {
+    content: "#{$mc-bp-sm}";
+  }
+
+  &--md:before {
+    content: "#{$mc-bp-md}";
+  }
+
+  &--lg:before {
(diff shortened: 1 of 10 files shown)


bernabe9:
@stephenp so now the container doesn't have a max-width? Is there another class to set a max width to the content?

stephenp:
@bernabe9 There is a max-width when you get a window wider than 1200px, but other than that, the container will be 100% of the window width with padding on the left and right.  Here is the documentation from design that outlines how they want the grid to behave:

image

"Margins" are the padding on the left and right of the grid
"Gutters" is the amount of padding between each column and the one next to it (implemented in code as 1/2 the "gutter" value they define on the inside of each column)
"Breakpoint" is the visual width of a box contained within a full 12-column row.  So for the max breakpoint of 1200px, the content is within a container with a max width of 1328 (1200px + 64px margin + 64px margin)

Issue #210, opened 2018-09-19
$mc-color-secondary needs to be $mc-color-gray-200


jstnjns:
![screen shot 2018-09-19 at 10 00 10 am](https://user-images.githubusercontent.com/249444/45768875-ded85780-bbf2-11e8-83ef-69169d2a5152.png)

As well, set $mc-color-secondary-hover to $mc-color-gray-300

stephenp:
Resolved

PR #211, opened 2018-09-20
Updating color variables

stephenp:
## Overview
Grays were out of order and corrected in figma comp, so updating here as well.

## Risks
Low, but colors may change if they were using the variables (intentionally / acceptable though)

## Changes
Grays can now be used gray-100 (darkest) to gray-600 (lightest) as one would expect.  Also cleaned up "slightly different" grays used in edge cases to instead use the closest standard grays available (secondary / tertiary variables).

## Issue
https://github.com/yankaindustries/mc-components/issues/210

### DIFF
--- a/src/styles/base/_variables.scss
+++ b/src/styles/base/_variables.scss
@@ -12,29 +12,28 @@ $mc-color-light:             #fff !default;
 $mc-color-dark:              #000 !default;
 
 // Colors
+$mc-color-gray-100:           #191c21 !default;
+$mc-color-gray-200:           #272c33 !default;
+$mc-color-gray-300:           #31333b !default;
+$mc-color-gray-400:           #596170 !default;
+$mc-color-gray-500:           #949aa4 !default;
+$mc-color-gray-600:           #dbdde1 !default;
+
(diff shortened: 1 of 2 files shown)


src/styles/base/_variables.scss
```diff
@@ -12,30 +12,29 @@ $mc-color-light:             #fff !default;
...
-$mc-color-secondary-hover:   #2c2c2e !default;
-$mc-color-secondary-active:  #1f1f21 !default;
+$mc-color-secondary:         $mc-color-gray-200 !default;
+$mc-color-secondary-hover:   $mc-color-gray-300 !default;
+$mc-color-secondary-active:  $mc-color-gray-100 !default;
 
-$mc-color-tertiary:          #75787f !default;
-$mc-color-tertiary-hover:    #90949d !default;
+$mc-color-tertiary:          $mc-color-gray-400 !default;
+$mc-color-tertiary-hover:    $mc-color-gray-500 !default;
```
jstnjns: eventually, we should prolly have a var for tertiary active as well

stephenp: Once it's comped!

PR #214, opened 2018-09-25
Dist reorg

stephenp:
## Overview
Reorganize the dist folder to work with inlining svg icons.  End users can now consume anything in `dist` without fear. (And now they don't need to access `src`).

## Risks
Medium - High. If anything outside of this repo is currently referencing the existing `src` files, they'll need to be updated. I did a "Find in project" for the main masterclass repo, but there's always a risk of missing a file / import.

## Changes
No visual changes.

## Issue
N/A


package.json
```diff
@@ -7,20 +7,21 @@
   "bugs": {
     "url": "https://github.com/yankaindustries/mc-components/issues"
   },
-  "main": "./dist/index.js",
-  "module": "./dist/index.js",
-  "style": "dist/styles/css/mc-components.css",
+  "main": "./dist/components/index.js",
```
stephenp: The new structure has everything in either:
- `/dist/components/`
- `/dist/styles/scss/`
- `/dist/styles/css/`

package.json
```diff
@@ -7,20 +7,21 @@
...
     "prebuild": "rimraf ./dist",
     "build": "yarn build:assets && yarn build:js && yarn build:styles",
     "build:assets": "svgr ./src/assets/icons --out-dir ./src/components/Icons",
-    "build:js": "babel ./src/components --out-dir ./dist --ignore '**/*.stories.js,**/*.test.js'",
-    "build:css": "node-sass ./src/styles/index.scss | postcss > ./dist/styles.css",
-    "build:css-vendor": "node-sass ./src/styles/libraries/_libraries.scss | postcss > ./dist/vendor.css",
-    "build:styles": "yarn build:css && yarn build:css-vendor",
-    "start": "start-storybook -p 9001 -c .storybook",
+    "build:js": "babel ./src/components --out-dir ./dist/components --ignore '**/*.stories.js,**/*.test.js'",
+    "build:scss": "NODE_ENV=development mkdir -p ./dist/styles/scss && postcss ./src/styles --base ./src/styles --dir ./dist/styles/scss",
```
stephenp: We now set the `node_env` variable to determine whether we're on production or development.

### On development
- postcss config `syntax` is set to post-css-scss and leaves scss "as is"
- postcss does not run `cssnano` 

### On production
- postcss syntax is set to `undefined` (basically, don't specify / use default)
- postcss css nano options enabled to minify output css.

PR #216, opened 2018-09-27
Legacy classnames

jstnjns:
## Overview
We are removing a legacy tile classnames as they are just conflicting with what's on prod, and should be cleaned out anyway.

## Risks
Low - We checked tile implementations in `masterclass` to see if there were any conflicts with removing these classes, they are using the local implementation anyway so there appears to be no issue...so long as we didn't miss any.

### DIFF
--- a/src/styles/components/_tile.scss
+++ b/src/styles/components/_tile.scss
@@ -8,7 +8,9 @@
     content: "";
   }
 
-  .mc-tile__content {
+  .mc-tile__content,
+  .background,
+  .content {
     position: absolute;
     left: 0;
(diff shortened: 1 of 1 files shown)


src/styles/components/_tile.scss
```diff
@@ -7,18 +7,9 @@
     padding-top: ($height / $width) * 100%;
     content: "";
   }
-
-  .mc-tile__content {
```
jstnjns: Unnecessary to belong in the mixin as the variables assigned above have no bearing on these values.

PR #220, opened 2018-10-01
Add onseek event

bernabe9:
## Overview
Add `onSeek` event to the video player

## Risks
None.


src/components/VideoPlayer/index.js
```diff
@@ -90,7 +91,12 @@ export default class VideoPlayer extends PureComponent {
 
   handlePlayerReady = () => {
     const {
-      onPlay, onPause, onVideoReady, onPlayerReady, pausescreenComponent,
+      onPlay,
```
jstnjns: could we start sorting these alphabetically?

PR #222, opened 2018-10-03
videoJS buffers bug

bernabe9:
## Overview
videoJS has a bug that doesn't load the audio buffer.

This code is going to reload the video if a buffer isn't working.

## Risks
Low


stephenp:
@nicastelo can you resolve the conflicts and add a "ready to merge" label when it's ready? Thanks!

PR #226, opened 2018-10-08
Reduce wildcard specificity of column selector

stephenp:
## Overview
- Tiles on homepage were not using the correct gutter sizes
- Account edit page was broken from using the `col` name in the middle of a class name.
- The overrides were imported in the wrong order, causing the overrides to not apply to vertical gutters

## Risks
Low - If any classes begin with `col`, they will inherit the bootstrap grid styles.  We will eventually use bootstrap's method of looping through each column type and breakpoint so collisions have a nil chance of happening.

## Changes
Changes columns behavior

### DIFF
--- a/src/styles/libraries/bootstrap-overrides/_grid.scss
+++ b/src/styles/libraries/bootstrap-overrides/_grid.scss
@@ -16,7 +16,8 @@
 }
 
 .col,
-[class*="col-"] {
+[class^="col-"],
+[class*=" col-"] {
   padding: $grid-gutter-width / 2;
 }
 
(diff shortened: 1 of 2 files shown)


src/styles/libraries/bootstrap-overrides/_grid.scss
```diff
@@ -16,7 +16,8 @@
 }
 
 .col,
-[class*="col-"] {
+[class^="col-"],
+[class*=" col-"] {
```
jstnjns: As we discussed, we will probably want to identify cols at each size at each breakpoint.  Just wanted anybody else seeing this to know what the eventual plan is.

stephenp: For now, we're sticking with this method for ease of implementation since the primary risk is affecting any classes that begin with `col` which is unlikely in the masterclass repo.  The correct implementation will be a bigger and more complex change in the SCSS which will render out each possible column name at each breakpoint and override the gutters.

PR #228, opened 2018-10-08
Adding `intent` value to HoverHandler

jstnjns:
## Overview
To avoid wild actions when cursor moves across a grid where tiles are listing for hover, we've added `intent` so that you can listen to hover only when the user has stopped moving the mouse.

## Risks
None - This is adding functionality, not changing any.

### DIFF
--- a/src/components/HoverHandler/index.js
+++ b/src/components/HoverHandler/index.js
@@ -15,6 +15,10 @@ export default class HoverHandler extends PureComponent {
     intent: false,
   }
 
+  componentWillUnmount () {
+    if (this.timer) this.timer = clearTimeout(this.timer)
+  }
+
   onEnter = () => {
     if (this.timer) this.timer = clearTimeout(this.timer)
 
@@ -32,7 +36,7 @@ export default class HoverHandler extends PureComponent {
     )
   }
 
-  onOut = () => {
+  onLeave = () => {
     if (this.timer) this.timer = clearTimeout(this.timer)
 
     this.setState({
@@ -53,7 +57,7 @@ export default class HoverHandler extends PureComponent {
     return (
       <span
         onMouseEnter={this.onEnter}
-        onMouseLeave={this.onOut}
+        onMouseLeave={this.onLeave}
       >
         {children({ hovering, intent })}
       </span>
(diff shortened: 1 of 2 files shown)


a-axton:
is it possible that the timeout could still exist if the component unmounts? should we clear the timer just incase?

jstnjns:
> is it possible that the timeout could still exist if the component unmounts? should we clear the timer just incase? eg, if the component is also a link to another page/route

Good call.  Updating

src/components/HoverHandler/index.js
```diff
@@ -2,34 +2,65 @@ import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 
 
+const DELAY = 300
```
bernabe9: What do you think of having a prop (with the default value 300) for the delay?

src/components/HoverHandler/index.js
```diff
@@ -2,34 +2,65 @@ import React, { PureComponent } from 'react'
...
     children: PropTypes.func.isRequired,
   }
 
   state = {
     hovering: false,
+    intent: false,
+  }
+
+  componentWillUnmount () {
+    if (this.timer) this.timer = clearTimeout(this.timer)
```
bernabe9: since the component is unmounting, do you need to assign the value of `clearTimeout` to `this.timer`?

```javascript
componentWillUnmount () {
  if (this.timer) clearTimeout(this.timer)
}
```

src/components/HoverHandler/index.js
```diff
@@ -2,34 +2,65 @@ import React, { PureComponent } from 'react'
...
+    const {
+      hovering,
+      intent,
+    } = this.state
 
     return (
-      <div
-        onMouseEnter={this.onEnter.bind(this)}
-        onMouseLeave={this.onLeave.bind(this)}
+      <span
```
bernabe9: just out for curiosity, why did you change it to a `span`?

src/components/HoverHandler/index.js
jstnjns: probably not

src/components/HoverHandler/index.js
jstnjns: thought about that, didn't really see a need

src/components/HoverHandler/index.js
jstnjns: so that this could wrap inline elements without having odd consequences or needing additional styling

PR #231, opened 2018-10-09
Subheader style fixed

juanlacruz:
## Overview
Subheader style was not taking into consideration due to a wrong className.

## Risks
Minor fix, no risks.

## Issue
Now the modal will use the defined subheader style.

### DIFF
--- a/src/styles/components/modal/_modal-header.scss
+++ b/src/styles/components/modal/_modal-header.scss
@@ -8,7 +8,7 @@
   text-align: center;
 }
 
-.mc-components-modal-header__subheader {
+.modal-header__subheader {
   color: rgba(#4d5768, 0.3);
   font-weight: 700;
   line-height: 2rem;


stephenp:
So I know this is going in to resolve the header styles missing in the masterclass repo modal usage, but looking at this PR: https://github.com/yankaindustries/mc-components/pull/169/files

@nicastelo made some changes so the modal class wouldn't conflict by renaming the parent class from `.modal` to `.mc-component-modal`.  Should the header follow the same convention, or do you think this is a unique enough name to not cause conflicts? Should it follow the same convention for ease of understanding in the code base?

juanlacruz:
@stephenp I think that as you said this is a unique enough name, and it shouldn't cause any conflicts. Anyway, if we want to keep the _.mc-component-modal_ name, we should change the Modal.js to include it.

PR #241, opened 2018-10-10
Building new <Modal /> component

jstnjns:
## Overview
The original Modal component was a little less flexible than what this component library is expected to be, so this is a replacement.  

## Risks
Medium - All modal implementations will need to be replaced.  We are not keeping around the legacy implementation to simplify adoption this time.

## Changes
The new biggest change to the Modal component is that it isn't very concerned with how it looks.  The only decision that has been made is where the close button lives (outside, top right, which I think to be reasonable to expect consistency in), as well as a semi-transparent black background.  Other than that, structural styling is decided by what other components you decide to use.

There are actually two components that are necessary, `Modal` and `ModalContent`.  Below you'll see why we need `ModalContent`.

This is code directly from the story that shows three ways you can (but are not limited to) implement the modal:
screen shot 2018-10-10 at 10 50 44 am

The outcome if clicking on all three buttons are as follows:

screen shot 2018-10-10 at 10 49 47 am
screen shot 2018-10-10 at 10 49 49 am
screen shot 2018-10-10 at 10 49 54 am

You'll see that the content within the `Modal` components are centered (vertically and horizontally) above the modal background layer.  Additionally, you'll notice that the grid system is used to position the content (in these cases, in other cases you can use whatever you like), and then we have the `ModalContent` component. `ModalContent`'s primary use is to designate which part of the modal is structural setup, and which part is actual content.  It is also assists in capturing focus so that users can use the escape key to exit.

The mechanism for showing and hiding the component largely remains the same, although the values are not required.  In the case that you want to display what appears to be a modal without controlling it (faux modal) it supports that as well.

Lastly, there used to be a `ModalPortal` component.  The way this was set up required you to have a mounting point for the modal.  We've removed this, and it simply appends to the modal to `document.body` and doesn't require a mounting point at all.

## Issue
https://masterclass-dev.atlassian.net/browse/WI-69

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -19,7 +19,7 @@ export default class Modal extends PureComponent {
     children: PropTypes.oneOfType([
       PropTypes.node,
       PropTypes.arrayOf(PropTypes.node),
-    ]),
+    ]).isRequired,
     className: PropTypes.string,
     show: PropTypes.bool,
 
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -1,0 +1,72 @@
+import React, { PureComponent } from 'react'
+import { createPortal } from 'react-dom'
+import PropTypes from 'prop-types'
+import cn from 'classnames'
+
+import IconClose from '../Icons/Close'
+
+
+export default class Modal extends PureComponent {
+  static propTypes = {
+    children: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
+    className: PropTypes.string,
+    show: PropTypes.bool,
+
+    onCloseClick: PropTypes.func,
+  }
+
+  onKeyDown = (event) => {
+    if (event.keyCode === 27) {
+      this.close()
+    }
+  }
+
+  close = () => {
+    const {
+      onCloseClick,
+    } = this.props
+
+    if (onCloseClick) {
+      onCloseClick()
+    }
+  }
+
+  render () {
+    const {
+      children,
+      className,
+      show,
+    } = this.props
+
+    if (!show) {
+      return null
+    }
+
+    const classes = cn({
+      [className]: className,
+      'mc-modal': true,
+    })
+
+    return createPortal(
+      <div
+        className={classes}
+        onKeyDown={this.onKeyDown}
+      >
+        <a
+          className='mc-modal__backdrop'
+          onClick={this.close}
+        >
+          <IconClose className='mc-modal__close' />
+        </a>
+
+        <div className='mc-modal__container'>
+          {children}
+        </div>
+      </div>,
+      document.body,
+    )
+  }
+}
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -6,6 +6,14 @@ import cn from 'classnames'
 import IconClose from '../Icons/Close'
 
 
+const ModalContext = React.createContext('modal')
+
+export const {
+  Provider,
+  Consumer,
+} = ModalContext
+
+
 export default class Modal extends PureComponent {
   static propTypes = {
     children: PropTypes.oneOfType([
@@ -20,17 +28,17 @@ export default class Modal extends PureComponent {
 
   onKeyDown = (event) => {
     if (event.keyCode === 27) {
-      this.close()
+      this.close('escape')(event)
     }
   }
 
-  close = () => {
+  close = source => (event) => {
     const {
       onCloseClick,
     } = this.props
 
     if (onCloseClick) {
-      onCloseClick()
+      onCloseClick(source, event)
     }
   }
 
@@ -39,6 +47,8 @@ export default class Modal extends PureComponent {
       children,
       className,
       show,
+
+      onCloseClick,
     } = this.props
 
     if (!show) {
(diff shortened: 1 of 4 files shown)


src/components/Modal/index.js
```diff
@@ -1,139 +1,72 @@
...
-          <Close
-            className={cnClose}
-            onClick={onClose}
-          /> }
-        <ClickOutside
-          divRef={this.container}
-          onClickOutside={this.onClickOutside}
+    const classes = cn({
+      [className]: className,
+      'mc-modal': true,
```
bernabe9: Can you use another classname?

We had a lot of issues with this classname because there are other modal implementations in masterclass with the same classname.

src/components/Modal/index.js
```diff
@@ -1,139 +1,72 @@
...
-    NONE: 'none',
-  }
 
+export default class Modal extends PureComponent {
   static propTypes = {
-    isOpen: bool.isRequired,
-    onClose: func.isRequired,
-    children: oneOfType([arrayOf(node), node]).isRequired,
-    shouldCloseOnEsc: bool,
-    shouldCloseOnClickOutside: bool,
```
bernabe9: Are we sure that we don't want to close the modal on click outside?

src/components/Modal/index.js
```diff
@@ -1,139 +1,72 @@
...
+      [className]: className,
+      'mc-modal': true,
+    })
+
+    return createPortal(
+      <div
+        className={classes}
+        onKeyDown={this.onKeyDown}
+      >
+        <a
```
bernabe9: use another html tag. `anchor` tag is used for hyperlinks

src/components/Modal/index.js
jstnjns: Part of integrating this will be replacing all other modal implementations.

src/components/Modal/index.js
jstnjns: Removing this property is not removing that functionality, it still remains.

src/components/ModalContent/index.js
```diff
@@ -0,0 +1,46 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import cn from 'classnames'
+
+
+export default class ModalContent extends PureComponent {
```
a-axton: what is the purpose of this component?

src/components/Modal/index.js
a-axton: I agree, we should use a different class name. What if we end up changing the structure of the react component and update the styles to match? If we use this class name we're locked into supporting and testing the old modals every time this gets updated

src/components/Modal/index.js
```diff
@@ -1,139 +1,72 @@
...
-                ])}
-                color='#000'
-                onClick={onClose}
-              />
-              {children}
-            </RoundedBox>
-          </div>
-        </ClickOutside>
-      </ModalPortal>
+          <IconClose className='mc-modal__close' />
```
a-axton: Modals don't always have an 'x' in the overlay, some of them have the 'x' inside the white rounded box and some don't have an 'x' at all

bernabe9: btw I think most of the modals that are implemented have the 'x' inside the modal.

Example:
screen shot 2018-10-10 at 6 46 58 pm

src/components/ModalContent/index.js
```diff
@@ -0,0 +1,46 @@
+import React, { PureComponent } from 'react'
```
a-axton: why can't this be inside of Modal? do we even need this component?

src/components/ModalContent/index.js
```diff
@@ -0,0 +1,46 @@
...
+export default class ModalContent extends PureComponent {
+  static propTypes = {
+    children: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
+    className: PropTypes.string,
+  }
+
+  constructor (props) {
```
bernabe9: you don't need to use the constructor.

```javascript
content = React.createRef()
```

Will work

src/components/Modal/index.js
jstnjns: I knew this one would come up.

So, with stripping out unnecessary parts of a modal, I battled with this one the most...but here's where I landed.  If the Modal component is opinionated on anything, it should be a consistent user experience (even if not a consistent design).  Where the close button exists should be something we can draw a line on.  Additionally, this is the least invasive place to put the close button for a designer.  If we cannot unify where a close button on a modal exists, we shouldn't be building a component library.

src/components/ModalContent/index.js
jstnjns: Dope, I dig that a lot more.  Thanks!

bernabe9:
We should consider the mobile modal. I tested and if the content is bigger than the screen we don't have space for the background and the "X" icon.


This is how it looks in this branch:
screen shot 2018-10-10 at 7 11 01 pm

This is how it looks in master:
screen shot 2018-10-10 at 7 12 20 pm

src/components/Modal/index.js
bernabe9: I think we will need the close button inside the modal for the mobile view. All the mobile design modals that I saw have the close button inside, just because there is no space in a mobile view to put it outside.

src/components/ModalContent/index.js
jstnjns: The summary of this PR explains `ModalContent` and why it exists, does it answer your questions?

a-axton: What I don't get is why it needs to be a separate component. You stated that it's to "designate which part of the modal is structural setup, and which part is actual content". What is the purpose of having a separate component instead of just putting all of this in Modal?

jstnjns: The purpose is so that structural markup between `<Modal>` and `<ModalContent>` is treated as part of the "backdrop."  This way, if you were to, as is in the story, wrap your content in a container and maybe size up the modal content using the grid system (see "small" example) the `.container`, `.row`, and `.col` divs don't block the click handler for the "outside click" to close the modal (or rather, fire the close click callback).

jstnjns: ^^ This is how we support all the different types of modal sizes and layouts without building it into the modal component itself.

a-axton: That's what I thought but I didn't see any click handlers closing the modal when the backdrop is clicked

src/components/Modal/index.js
jstnjns: i have the X hovering atop the content, so in the case that a modal is too large the close button is still displayed.  i can run this by design

PR #253, opened 2018-10-17
Minor tweaks for login modals

jstnjns:
## Overview
Updates to the component library realized when implementing for login modals

### DIFF
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -118,7 +118,7 @@ export default class Input extends PureComponent {
                 htmlFor={name}
                 className='mc-form-input__label'
               >
-                {label} {error}
+                {error || label}
               </label>
             }
           </div>


src/components/InputField/index.js
```diff
@@ -12,10 +12,14 @@ const FIELD_PROP_TYPE = {
 
 const renderInput = ({
   input,
+  meta: {
```
missx: Can yo add this as a prop type also?

src/components/Input/index.js
```diff
@@ -113,12 +113,12 @@ export default class Input extends PureComponent {
...
             />
 
-            {label &&
+            {(error || label) &&
               <label
                 htmlFor={name}
                 className='mc-form-input__label'
               >
-                {label}
+                {label} {error}
```
bernabe9:
```suggestion
                {error || label}
```

src/components/InputField/index.js
jstnjns: yes, good call

src/components/Input/index.js
jstnjns: I, initially, had exactly that.  Errors that come from the server are actually incredibly useful in the format they are in:

```json
{ "email": "can't be blank" }
```

leads well to be rendered like `Email can't be blank`

jstnjns: My idea fails if the label isn't the name of the field.

`Tell us how you would describe yourself can't be blank` fails...  Changing.  Thanks!

Issue #257, opened 2018-10-23
Clean up unused or unnecessary variables


a-axton:
There are lots of unused variables like `$default-btn-line-height` and some color variables that I think we can get rid of like `$mc-color-secondary` and `$mc-color-tertiary` which are just re-assigned `mc-color-x` variables

a-axton:
https://github.com/yankaindustries/mc-components/pull/259

PR #261, opened 2018-10-29
Cleanup unused mixins and vars

a-axton:
# DO NOT MERGE
Until this PR is approved and merged in: https://github.com/yankaindustries/masterclass/pull/7613

## Overview
Original goal was to remove some extra vars in variables.scss but I ended up doing much more.
* remove unused vars
* moved mixins and functions dirs in to base dir
* remove bootstrap vars, moved required grid vars into our vars folder (should get rid of that pesky assert ascending warning)
* removed mixins that were unused or that seemed pointless (transition, box-shadow)
* removed button mixins, they were only being used on the facebook button

I've created this branch to update all of the missing mixins/vars in the masterclass repo https://github.com/yankaindustries/masterclass/pull/7613

## Risks
Medium - merging into this repo
High - versioning into masterclass repo
* Some of the variables I removed are being used in the client stylesheet and others. They are only being used once or twice in each place.
* We'll have to replace buttons in client with mc-component buttons, they are using a lot of the vars/mixins I removed. We need to do this anyways so it's a good excuse
* Some of the mixins I removed are being used in the repo (should be easy to replace with regular css)

## Issue
https://github.com/yankaindustries/mc-components/issues/257


src/styles/base/_base.scss
```diff
@@ -1,3 +1,3 @@
 @import "variables";
-@import "mixins";
-@import "helpers";
+@import "functions/functions";
```
jstnjns: since these are single file dirs, can we just get rid of the dirs?

PR #266, opened 2018-10-29
Updating typography to work with seo page and future pages

stephenp:
## Overview
The new SEO page relies heavily on typography and is the perfect test arena for mc-components typography. While working to build out the page, I adjusted the typography we have in mc-components to work with this page.  This is an ongoing learning experience and we will continue to update type styles in the future, but this is a good starting point.

## Risks
No risk for mc-components, but medium risk for the masterclass repo. We are using existing type styles in a few locations in the masterclass repo (outlined below) - these areas will need to be verified as acceptable before versioning.

## Changes
Type updates - font sizing, responsive type changes, new type headings

## Issue
N/A

## Changes
- Base font size now set to 16px and 1.5 line height
- `h1` changed from `32px` on desktop to `48px`
- `h2` changed from `24px` on desktop to `32px`
- `h3` changed from `20px` on desktop to `28px`
- `h4` changed from `14px` on desktop to `24px`
- `h5` changed from `12px` on desktop to `20px`

Type scales down on a sliding scale among these values: 

`48px, 32px, 24px, 20px, 16px, 12px`

So if `h1` is `48px` on desktop, on tablet it will be `32px`, mobile it will be `24px`.

### DIFF
--- a/src/foundation/type/index.stories.js
+++ b/src/foundation/type/index.stories.js
@@ -74,6 +74,14 @@ storiesOf('foundation|Type', module)
         </div>
 
         <div className='align-items-center example--section'>
+          <h5 className='mc-text-h5'>Readable</h5>
+          <p className='mc-text-intro'>Online classes taught by the world&#39;s greatest minds. Now get
+           unlimited access to all classes.
+          </p>
+          <p className='mc-text--muted mc-text--monospace'>.mc-text-readable</p>
+        </div>
+
+        <div className='align-items-center example--section'>
           <h5 className='mc-text-h5'>Legal</h5>
           <p className='mc-text-legal'>Copyright 2018 MasterClass</p>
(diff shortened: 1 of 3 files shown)


src/foundation/type/index.stories.js
```diff
@@ -73,6 +73,14 @@ storiesOf('foundation|Type', module)
           <p className='mc-text--muted mc-text--monospace'>.mc-text-intro</p>
         </div>
 
+        <div className='align-items-center example--section'>
+          <h5 className='mc-text-h5'>Readable</h5>
+          <p className='mc-text-intro'>Online classes taught by the world&#39;s greatest minds. Now get
```
jstnjns: uses `mc-text-intro`

not technically a problem because they share the same styles, but that might not always be the case

Issue #275, opened 2018-11-01
Consistent invert styles


stephenp:
After building `invert` helpers for each individual component like `mc-text--invert`, `mc-input-checkbox--invert`, `mc-form-input--invert`, `mc-input-radio--invert`, `mc-form-textarea--invert` and more...with the new finalized buttons, I've added a helper class called `mc-invert` that can be added on the parent wrapper.

All previous implementations of invert styles should be updated to have this same treatment for consistency. (You should be able to add a `mc-invert` class to the parent of any wrapper and all children get updated to have the invert style without having to set it manually on each element).

PR #282, opened 2018-11-06
Invert updates

stephenp:
# DO NOT MERGE

## Overview
We have moved from having the `--invert` modifier on each element type to instead wrapping elements at the container level with `mc--invert` instead. This PR adapts those changes, and will need to have components / style classes updated in the main masterclass repo.

### Classes updated
The following classes have been removed from the code base.  The proper way to introduce inverted styles is to add a css class of `mc--invert` to the parent wrapper.
  - mc-form-textarea--invert
  - mc-input-checkbox--invert
  - mc-text--invert
  - mc-form-input--invert
  - mc-input-radio--invert
The form invert behavior has also changed. Forms were initially built assuming a white background as the default, with a black background as being invert. To remain consistent with the rest of the invert behaviors, forms will now default to assuming they're on a black background, and the invert styles will be for use on a white background.

## Risks
High.

## Changes
No visible changes as long as classes are switched over properly.

## Issue
https://github.com/yankaindustries/mc-components/issues/275

### DIFF
--- a/src/styles/components/forms/textarea/_default.scss
+++ b/src/styles/components/forms/textarea/_default.scss
@@ -4,8 +4,8 @@ $scrollbarWidth: 7px !default; // Browser scrollbar width
 .mc-form-textarea {
   position: relative;
   width: 100%;
-  background: $mc-color-light;
-  border: 1px solid $mc-color-gray-600;
+  background: $mc-color-dark;
+  border-color: $mc-color-gray-300;
   border-radius: 4px;
   font-size: 15px;
   line-height: 1.5;
@@ -58,8 +58,6 @@ $scrollbarWidth: 7px !default; // Browser scrollbar width
 
   &--focus,
   &--modified {
-    // border-color: $mc-color-gray-400;
-
     .mc-form-textarea {
       &__textarea {
         padding: 20px $textarea-padding $textarea-padding;
@@ -73,7 +71,6 @@ $scrollbarWidth: 7px !default; // Browser scrollbar width
       &__label {
         font-size: 11px;
         transform: translateY(4px);
-        background: $mc-color-light;
         width: calc(100% - (2 * #{$textarea-padding}) - #{$scrollbarWidth});
         transition:
           font 320ms ease 0ms,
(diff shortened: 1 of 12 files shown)


jstnjns:
It appears that Textarea isn't styled properly.  I don't know if that is introduced by this PR, or existing.

screen shot 2018-11-13 at 11 51 52 am

src/styles/components/forms/textarea/_default.scss
```diff
@@ -4,8 +4,8 @@ $scrollbarWidth: 7px !default; // Browser scrollbar width
 .mc-form-textarea {
   position: relative;
   width: 100%;
-  background: $mc-color-light;
-  border: 1px solid $mc-color-gray-600;
+  background: $mc-color-dark;
+  border-color: $mc-color-gray-300;
```
jstnjns: Looks like here, there's no width set.

src/styles/components/forms/textarea/_default.scss
```diff
@@ -4,8 +4,8 @@ $scrollbarWidth: 7px !default; // Browser scrollbar width
 .mc-form-textarea {
   position: relative;
   width: 100%;
-  background: $mc-color-light;
-  border: 1px solid $mc-color-gray-600;
+  background: $mc-color-dark;
```
jstnjns: This should probably match the input background.

src/styles/components/forms/textarea/_modifiers.scss
```diff
@@ -13,17 +13,13 @@
   }
 }
 
-.mc-form-textarea--invert {
-  background: $mc-color-dark;
-  border-color: $mc-color-gray-300;
+@at-root .mc--invert .mc-form-textarea {
```
jstnjns: Per our discussion, let's get rid of `@at-root` here and everywhere else

src/styles/typography/_modifiers.scss
```diff
@@ -17,7 +17,7 @@
   &--airy { letter-spacing: 0.24em !important; }
 
   // Colors
-  &--invert { color: $mc-color-dark; }
+  @at-root .mc--invert & { color: $mc-color-dark; }
```
jstnjns: Remove `&`

PR #295, opened 2018-11-14
Adding dots to carousel

stephenp:
## Overview
Slick slider has a "dots" option to add thumbnail dots to a slider. This PR adds some styles to override their default, as well as passes along the option as a boolean.

## Risks
None...I don't know if this is the right way to do this, or if I should pass a bunch of options, including custom markup like the `arrows` example. Will have @jstnjns weigh in on this.

## Changes
dots

## Issue
N/A

### DIFF
--- a/src/components/Carousel/index.js
+++ b/src/components/Carousel/index.js
@@ -72,6 +72,7 @@ export default class Carousel extends PureComponent {
     children: PropTypes.node,
     className: PropTypes.string,
     controls: PropTypes.bool,
+    dots: PropTypes.bool,
     fadeEdges: PropTypes.bool,
     focusOnSelect: PropTypes.bool,
     sliderRef: PropTypes.func,
@@ -86,6 +87,7 @@ export default class Carousel extends PureComponent {
     autoPlay: false,
     centered: false,
     controls: false,
+    dots: false,
     fadeEdges: false,
     focusOnSelect: false,
     loop: false,
@@ -102,6 +104,7 @@ export default class Carousel extends PureComponent {
       className,
       centered,
       controls,
+      dots,
       fadeEdges,
       loop,
       sliderRef,
@@ -149,6 +152,7 @@ export default class Carousel extends PureComponent {
           slidesToScroll={scrollCount}
           slidesToShow={showCount}
           infinite={loop}
+          dots={dots}
           {...arrows}
           {...restProps}
         >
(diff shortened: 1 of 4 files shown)


src/components/Carousel/index.js
```diff
@@ -149,6 +152,7 @@ export default class Carousel extends PureComponent {
           slidesToScroll={scrollCount}
           slidesToShow={showCount}
           infinite={loop}
+          dots={dots}
```
stephenp: Should this be `{...dots}` so we can pass in custom options like `dotsClass` in addition / make our own adjustments to dots?

src/styles/components/_carousel.scss
```diff
@@ -1,5 +1,39 @@
 $mc-carousel-centered-padding: 100px !default;
 
+.slick-dots {
+  /* stylelint-disable */
+  display: flex !important;
```
stephenp: I hate having to use `!important` here, but slick slider inlines `style="display: block;"` when the dots should be visible.

src/styles/components/_carousel.scss
```diff
@@ -1,5 +1,39 @@
 $mc-carousel-centered-padding: 100px !default;
 
+.slick-dots {
```
stephenp: Passing `dotsClass` to slick slider would allow us to set a custom class here.

src/styles/components/_carousel.scss
```diff
@@ -1,5 +1,39 @@
...
+    cursor: pointer;
+    transition: background 250ms ease;
+    outline: 0;
+
+    &:hover {
+      background: $mc-color-gray-300;
+    }
+  }
+
+  .slick-active button {
```
stephenp: I'm not sure if we can override slick's default class for active or not. [Documentation is here](http://kenwheeler.github.io/slick/) for slick slider.

src/components/Carousel/index.js
jstnjns: Actually, the changes to this file to add passing along of dots isn't necessary.  `{...restProps}` below passes all props.  The only reason we pass the props above is because the names are changed.

src/styles/components/_carousel.scss
jstnjns: If you want to change the name we can, otherwise this is fine too.  If we don't change the name, maybe we would want to wrap this in something like `.mc-carousel` so that we don't affect other instances that might have dots?

src/styles/components/_carousel.scss
jstnjns: Same comment here re: wrapping class in namespace.

PR #303, opened 2018-11-19
Background color

jstnjns:
## Overview
The ability to add color to a Background component.  This will simplify how we apply common backgrounds to sections

## Risks
None

## Changes
screen shot 2018-11-18 at 10 46 28 pm
screen shot 2018-11-18 at 10 46 40 pm

### DIFF
--- a/src/components/Background/index.js
+++ b/src/components/Background/index.js
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
+import cn from 'classnames'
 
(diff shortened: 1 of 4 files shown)


jstnjns:
@stephenp do you think the white background should automatically apply invert class (when it's ready)?  i can't think of a time when we wouldn't want to invert the styles with a white background

stephenp:
After thinking more on this - I think the colors should be set to what they're defined to and ignore "mc-invert" at the parent level. If things break, you need to go back in and invert manually.  It's a rare edge case that we can resolve when we run into it.

PR #305, opened 2018-11-20
keyboard handlers on VidepPlayer component

garura:
## Overview
https://masterclass-dev.atlassian.net/wiki/spaces/RET/pages/62750750/In-video+Keyboard+Controls

## Risks
Breaking the video player in some unexpected way

## Changes
The VideoPlayer component will now have keydown listeners to pause/play on spacebar, skip with side arrow keys, and increase/decrease volume with up/down arrow keys. These listeners should only be active while the video player is focused.

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -171,6 +171,78 @@ export default class VideoPlayer extends PureComponent {
 
   resumeVideo = () => { this.video.play() }
 
+  handleKeyDown = (e) => {
+    const { target } = e
+    const selectingVolume = target.className.indexOf('vjs-volume-bar') > -1 ||
+      target.className.indexOf('vjs-volume-menu-button') > -1
+
+    if (selectingVolume) {
+      // trying to prevent these listeners from affecting accessibility controls
+      return
+    }
+
+    switch (e.key) {
+      case 'ArrowLeft': {
+        e.preventDefault()
+        this.playerSkipBackwards()
+        break
+      }
+      case 'ArrowRight': {
+        e.preventDefault()
+        this.playerSkipForward()
+        break
+      }
+      case ' ': {
+        e.preventDefault()
+        this.playPausePlayer()
+        break
+      }
+      case 'ArrowUp': {
+        e.preventDefault()
+        this.increasePlayerVolume()
+        break
+      }
+      case 'ArrowDown': {
+        e.preventDefault()
+        this.decreasePlayerVolume()
+        break
+      }
+      default:
+    }
+  }
+
+  increasePlayerVolume = () => {
+    const currentVolume = this.video.volume()
+    const newVolume = Math.min(currentVolume + 0.05, 1)
+    this.video.volume(newVolume)
+  }
+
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -171,6 +171,78 @@ export default class VideoPlayer extends PureComponent {
 
   resumeVideo = () => { this.video.play() }
 
+  handleKeyDown = (e) => {
```
jstnjns: if you wanted to, you could pull `target` and `preventDefault` here since that's all that's used.  not required, just a suggestion

src/components/VideoPlayer/index.js
```diff
@@ -171,6 +171,78 @@ export default class VideoPlayer extends PureComponent {
...
+        this.decreasePlayerVolume()
+        break
+      }
+      default:
+    }
+  }
+
+  increasePlayerVolume = () => {
+    const currentVolume = this.video.volume()
+    const newVolume = Math.min(currentVolume + 0.05, 1)
```
jstnjns: could we pull the `0.05` out into a constant instead of having magic numbers in the code?  (here and `decreasePlayerVolume`

src/components/VideoPlayer/index.js
```diff
@@ -171,6 +171,78 @@ export default class VideoPlayer extends PureComponent {
...
+      case 'ArrowDown': {
+        e.preventDefault()
+        this.decreasePlayerVolume()
+        break
+      }
+      default:
+    }
+  }
+
+  increasePlayerVolume = () => {
```
jstnjns: please remove `Player` from method names, `increaseVolume` should be 👌

src/components/VideoPlayer/index.js
```diff
@@ -171,6 +171,78 @@ export default class VideoPlayer extends PureComponent {
...
+    if (this.video.paused()) {
+      this.video.play()
+    } else {
+      this.video.pause()
+    }
+  }
+
+  playerSkipForward = () => {
+    const currentTime = Math.floor(this.video.currentTime())
+    const newTime = Math.min(this.video.duration(), currentTime + 5)
```
jstnjns: same as volume increase, please move time change to constant

src/components/VideoPlayer/index.js
garura: that's cool

garura: looks like destructuring `preventDefault` out doesn't work unless I bind it to e again

src/components/VideoPlayer/index.js
```diff
@@ -171,6 +174,78 @@ export default class VideoPlayer extends PureComponent {
 
   resumeVideo = () => { this.video.play() }
 
+  handleKeyDown = (e) => {
+    const { target, key } = e
+    const selectingVolume = target.className.indexOf('vjs-volume-bar') > -1 ||
```
a-axton: just an FYI, another way to do this is `target.classList.contains('vjs-volume-bar')`

PR #307, opened 2018-11-21
Updating positioning for tilecaption

stephenp:
## Overview
We need more control over tile caption positioning.  This adds x and y axis control for how the caption aligns.

## Risks
TileCaption previously did not have x and y values, so any place that TileCaption is currently used needs to be updated.  It looks like it's only used at the following location in the main repo right now:

`app/javascript/packs/client/features/promotions/BogoPage/BogoPage.js`

This will be a **breaking change** for anyone that is currently using `TileCaption` or the styles associated with it in their project.  This should be communicated to everyone.

#### Old styles
image

#### New Styles:
image

## Issue
N/A

### DIFF
--- a/src/components/TileCaption/index.js
+++ b/src/components/TileCaption/index.js
@@ -10,15 +10,20 @@ export default class TileCaption extends PureComponent {
     ]),
     className: PropTypes.string,
     position: PropTypes.oneOf([
-      'left bottom',
-      'center bottom',
-      'left below',
-      'center below',
+      'x-left y-center',
+      'x-left y-bottom',
+      'x-left y-below',
+      'x-center y-center',
+      'x-center y-bottom',
+      'x-center y-below',
+      'x-right y-center',
+      'x-right y-bottom',
+      'x-right y-below',
     ]),
   }
 
   static defaultProps = {
-    position: 'left bottom',
+    position: 'x-left y-bottom',
   }
 
   render () {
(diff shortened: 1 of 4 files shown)


src/styles/components/_tile.scss
```diff
@@ -207,29 +209,67 @@
...
+
+  &--y-below {
+    .mc-tile-caption__content {
+      align-self: flex-end;
+      transform: translateY(100%);
+      background: rgba($mc-color-dark, 1);
+    }
+  }
+
+  // &--bottom {
```
jstnjns: kill this

src/components/TileCaption/index.js
```diff
@@ -28,13 +34,19 @@ export default class TileCaption extends PureComponent {
       position,
     } = this.props
 
-    const positionClasses =
-      position.split(' ').map(pos => `mc-tile-caption--${pos}`)
+    // eslint-disable-next-line prefer-const
+    let [x, y] = position.split(' ')
```
jstnjns: Wow, nicely done!

src/components/TileCaption/index.js
```diff
@@ -28,13 +34,19 @@ export default class TileCaption extends PureComponent {
...
+
+    // Support shorthand for 'center' => 'center center'
+    if (x === 'center' && !y) {
```
jstnjns: Clever, you are so good at this!

PR #309, opened 2018-11-23
Select component

nicastelo:
## Overview
Add `<Select>` component to mc-components since it's being used at least twice in the core codebase

## Risks
None

## Changes
screen shot 2018-11-23 at 9 26 56 am


src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+      placeholder,
+      onChange,
+      onFocus,
+      defaultValue,
+    } = this.props
+
+    return (
+      <Select
+        options={options}
+        styles={this.styles}
```
jstnjns: What's this?

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+        placeholder={placeholder}
+        onChange={onChange}
+        onFocus={onFocus}
+        className='mc-select'
+        classNamePrefix='mc-select__wrapper'
```
jstnjns: I believe this can be `'mc-select'`, as it is the prefix for all classes WITHIN the component.

screen shot 2018-11-26 at 12 11 18 pm

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+        isSearchable={false}
+        components={{
+          DropdownIndicator: this.renderDropdownIndicator(),
+          ...this.props.components,
```
jstnjns: I feel like, as a component library, we should probably be more decisive about how the select is rendered, and not allow the user to change how a `Select` is rendered (to this degree).

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
+import React, { PureComponent } from 'react'
+import { string, bool, array, func, object } from 'prop-types'
+import Select, { components } from 'react-select'
+
+export default class SelectInput extends PureComponent {
+  static propTypes = {
+    onChange: func.isRequired,
```
jstnjns: Are we not treating this as a controlled component?  I see that we are not passing `value`.  How well does this integrate with a `redux-forms` implementation?

src/styles/components/_select.scss
```diff
@@ -0,0 +1,102 @@
+.mc-select {
+  $grey-1: #cbcdd0;
```
jstnjns: We have a LOT of gray variables already, we should be using those.

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+    return (
+      <Select
+        options={options}
+        styles={this.styles}
+        placeholder={placeholder}
+        onChange={onChange}
+        onFocus={onFocus}
+        className='mc-select'
+        classNamePrefix='mc-select__wrapper'
+        isSearchable={false}
```
jstnjns: Isn't passing the prop...

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+  handleCloseMenu = () =>
+    this.setState({ isOpen: false })
+
+  renderDropdownIndicator = () => {
+    const { isOpen } = this.state
+    const DropdownIndicator = props =>
+      components.DropdownIndicator &&
+        <components.DropdownIndicator {...props}>
+          <Arrow rotate={isOpen} />
```
jstnjns: It appears that we might get this for free in the `react-select` documentation: https://github.com/JedWatson/react-select#controllable-props

`props.selectProps.menuIsOpen`

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+  handleOpenMenu = () =>
+    this.setState({ isOpen: true })
+
+  handleCloseMenu = () =>
+    this.setState({ isOpen: false })
+
+  renderDropdownIndicator = () => {
+    const { isOpen } = this.state
+    const DropdownIndicator = props =>
+      components.DropdownIndicator &&
```
jstnjns: Is there ever a case where this doesn't exist?

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+  handleOpenMenu = () =>
+    this.setState({ isOpen: true })
+
+  handleCloseMenu = () =>
+    this.setState({ isOpen: false })
+
+  renderDropdownIndicator = () => {
+    const { isOpen } = this.state
+    const DropdownIndicator = props =>
```
jstnjns: Should this be defined outside of the render method?  It doesn't need to be redefined on every render, right?

src/components/Select/index.js
```diff
@@ -0,0 +1,89 @@
...
+      components.DropdownIndicator &&
+        <components.DropdownIndicator {...props}>
+          <Arrow rotate={isOpen} />
+        </components.DropdownIndicator>
+    return DropdownIndicator
+  }
+
+  render () {
+    const {
```
jstnjns: Since these are all the same name as the props being passed to `<Select />` below, could we simply expand all props?

```
<Select
  className='mc-select'
  moreManualProps='here'
  {...this.props}
/>
```

src/styles/components/_select.scss
```diff
@@ -0,0 +1,102 @@
...
+
+        &:active {
+          background-color: $grey-1;
+        }
+      }
+    }
+  }
+
+  // Mobile only
+  @media (max-width: 768px) {
```
jstnjns: There is a variable for this.

src/components/Select/index.stories.js
```diff
@@ -0,0 +1,37 @@
...
+    value: 4,
+    label: 'Option D',
+  },
+  {
+    value: 5,
+    label: 'Option Z',
+  },
+]
+
+storiesOf('components|Select', module).add('Select', withProps(Select)(() => (
```
jstnjns: This is a form element, should probably put it in that category

src/components/Select/index.js
nicastelo: That's because I need one level more of specificity in order to be able to override the default style.
That way I can do 

```
.mc-select .mc-select__wapper {}
```

Using
```
.mc-select .mc-select {}
```
looks weird

src/components/Select/index.js
nicastelo: We are using it in this scenario, in wich we need to override the default option component to add the check marks.

If we wanted to add feature with a closer component API we would need to re-implement the entire component instead of reusing this one.

screen shot 2018-11-29 at 1 53 21 pm

stephenp:
Updated to the latest version of mc-components `develop` branch.  @jstnjns can you take another look at this and provide some updated comments based on @nicastelo's latest changes from Nov?

stephenp:
Design has updated their select box implementation that they'd like to have implemented, which looks like this:

image

We will most likely be revisiting this in the near future and implementing their final version once it's done.  I'm going to close this PR, but we will reference it while building design's final forms to see what we can reuse.

Thank you @nicastelo for this first iteration, we'll build on it from here on out!

stephenp:
Oh, you can poke around here for their current ideation process:

https://www.figma.com/file/KtHjhreKwXt8aei3hlVd3c/Components-2.0-test-(sean-kesterson)?node-id=3606%3A451

Issue #289, opened 2018-11-09
Display function prop arguments in documentation.


stephenp:
Would be helpful to know what arguments a function prop is being called with.

https://ourmasterclass.slack.com/archives/GBC5JV6HF/p1541801364023600

PR #330, opened 2019-01-02
Improving documentation for <VideoPlayer />

jstnjns:
## Overview
Better documentation for the `<VideoPlayer />` component.  This includes prop definitions, function prop arguments, and demos.

## Risks
None

## Changes
screen shot 2018-12-31 at 11 45 20 am

screen shot 2019-01-01 at 9 28 55 pm

screen shot 2019-01-01 at 9 28 18 pm

 

## Issue
#289

### DIFF
--- a/src/components/TileVideo/index.js
+++ b/src/components/TileVideo/index.js
@@ -25,7 +25,7 @@ export default class TileVideo extends PureComponent {
     autoPlay: false,
     controls: false,
     loop: false,
-    muted: true,
+    muted: false,
   }
 
   constructor (props) {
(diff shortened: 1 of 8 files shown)


src/components/TileVideo/index.js
```diff
@@ -25,7 +25,7 @@ export default class TileVideo extends PureComponent {
     autoPlay: false,
     controls: false,
     loop: false,
-    muted: false,
+    muted: true,
```
stephenp: Intended change? All videos now muted by default?

jstnjns: ha, nice catch.  i must have set that while i was coding so that i could continue to listen to music 😂

jstnjns: jk, this is intentional.  thought this was on the `VideoPlayer`, but yes... `VideoTile` should be muted by default

PR #331, opened 2019-01-02
Transpile code to CommonJS

bernabe9:
## Overview
Adding `transform-es2015-modules-commonjs` to transpile our published code to common js

## Risks
None

## Why
A public library should be distributed in CommonJS


jstnjns:
not that it should change anything, but have you tested this by linking it in the masterclass repo locally?

bernabe9:
@jstnjns yes I have tested it in masterclass repo 👍

PR #340, opened 2019-01-11
Spacing helpers

stephenp:
## Overview
Adds spacing helpers for margins and padding.  It seems like overkill, but after building a few features, we've discovered the majority of CSS needing to be written is for these variables (margins and padding).  These helper classes will allow users to quickly add the spaces they need with some helper classes, while keeping the look consistent with the type scale.  We shouldn't have "custom" values for margins and padding anymore once this is in.

## Risks
None - updated previous "beta" examples where applicable, these are all new classes.

## Changes
Adds ability to use class names like `mc-mb-5` to add a `margin-bottom` value of "5" from the type scale. 

## Issue
N/A

### DIFF
--- a/src/styles/helpers/_spacing.scss
+++ b/src/styles/helpers/_spacing.scss
@@ -1,0 +1,83 @@
+// These are modeled after the bootstrap helper classes.
+
+// The premise is mc-m is the base for margins and mc-p is
+// the base for padding.  Here are some examples:
+
+// mc-m-6 = margin "6 scale" all around
+// mc-mt-6 = margin-top "6"
+// mc-mr-6 = margin-right "6"
+// mc-mb-6 = margin-bottom "6"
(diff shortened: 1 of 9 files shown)


src/styles/helpers/_spacing.scss
```diff
@@ -0,0 +1,83 @@
+// These are modeled after the bootstrap helper classes.
+
+// The premise is mc-m is the base for margins and mc-p is
+// the base for padding.  Here are some examples:
+
+// mc-m-6 = margin "6 scale" all around
+// mc-mt-6 = margin-top "6"
+// mc-mr-6 = margin-right "6"
+// mc-mb-6 = margin-bottom "6"
+// mc-ml-6 = margin-left "6"
```
jstnjns: can we add `.mc-mx-#` and `.mc-my-#`?

PR #341, opened 2019-01-11
Updating button examples with icons

stephenp:
## Overview
Updating examples 

## Risks
None

## Changes
Make sure any button items are vertically centered (like icons) - Also update documentation to show the correct margins on button icons.

## Issue

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -35,7 +35,11 @@ const Button = ({
       {...props}
     >
       {Boolean(Icon) && Icon}
-      <span>
+      <span className='
+        d-inline-flex
+        align-items-center
+        justify-content-center
+      '>
         {children}
       </span>
       {loading && <Loader className='loader' />}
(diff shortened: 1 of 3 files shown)


src/components/Button/index.js
```diff
@@ -35,7 +35,11 @@ const Button = ({
       {...props}
     >
       {Boolean(Icon) && Icon}
-      <span>
+      <span className='
+        d-inline-flex
```
stephenp: Is this the right way to do this?

jstnjns:
```
<span className={`
  d-inline-flex
  ...
`}>

File: src/components/Button/index.stories.js [hunk 2]
```diff
@@ -126,14 +126,13 @@ storiesOf('components|Button', module)
...
-                  <IconArrow />
+                  <IconArrow className='mc-ml-2' />
                 </Button>
               </div>
 
               <div className='col-auto'>
                 <Button secondary>
-                  <IconLock />
+                  <IconLock className='mc-mr-2' />
                   {' '}
```
jstnjns: can remove, i believe

PR #346, opened 2019-01-15
Adding helpers to limit total lines of text

stephenp:
## Overview
There was a request to add helpers that will limit the total lines of text shown, automatically hiding anything exceeding x length.  This PR adds two helpers, one for a 2-line limit, and one for a 3-line limit.  They can be used by adding the class `mc-text--2-lines` or `mc-text--3-lines` to any `mc-text` class, or any text tag used in HTML.

This PR also adds three new variables: `$mc-lh-sm`, `$mc-lh-lgm`, and `$mc-lh-lg`, which correspond to the line heights `1.25`, `1.5`, and `1.75` respectively.  This change was made so it's more obvious to developers working on mc-components that these line-heights are important and should be investigated further before making changes as they'll affect the ability of these new line-limiting classes to do their job.

## Risks
We will now need to update these tags if the line height values ever change, or if new tags are updated with custom line-height values.  I would rate this as a moderate risk - it has a high chance of happening, but if it does, it's not a show-stopping bug.

## Changes
image

## Issue
N/A


src/styles/typography/_modifiers.scss
```diff
@@ -23,6 +23,40 @@
...
+
+    // Default body / anything not defined
+    max-height: #{2 * $mc-lh-md}em;
+
+    // Headings
+    &[class^="mc-text-h"] {
+      max-height: #{2 * $mc-lh-sm}em;
+    }
+
+    // Headings
```
jstnjns: not a show stopper, but wrong label here and below

PR #351, opened 2019-01-21
Clean memory leaks when video player unmounts

bernabe9:
## Overview
Clean listeners and remove the DOM element once the video player unmounts.

## Risks
Medium -

In the playlist watch page we are manually destroying the video player, so we will need to remove that line.

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -95,10 +95,10 @@ export default class VideoPlayer extends PureComponent {
 
   componentWillUnmount () {
     // clean listeners
-    this.video.on('play', () => {})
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -93,6 +93,16 @@ export default class VideoPlayer extends PureComponent {
     }
   }
 
+  componentWillUnmount () {
+    // clean listeners
+    this.video.on('play', () => {})
```
jstnjns: Should this use `this.video.off('play')` instead?

bernabe9: yeah, good catch. It was fixed!

PR #355, opened 2019-01-28
Logged in Homepage POC

stephenp:
## Overview
Proof of concept for the new Logged in Homepage

## Risks
- Need to make sure `<Badge>` is updated in header in MC repo because content is passed into that component differently now.

## Changes
- Rounded corners no more (https://ourmasterclass.slack.com/archives/CB74RPL82/p1547754861001700)

## Issue
TBD


src/components/VideoPlayer/index.js
```diff
@@ -113,6 +113,8 @@ export default class VideoPlayer extends PureComponent {
       pausescreenComponent,
     } = this.props
 
+    console.log(this.video)
```
jstnjns: kill it

src/styles/base/mixins/_ratios.scss
```diff
@@ -4,48 +4,10 @@
...
 @mixin aspect-ratio($width, $height) {
-  position: relative;
   &:before {
     display: block;
     width: 100%;
     padding-top: ($height / $width) * 100%;
     content: "";
   }
-
-  /* Todo - Make sure this is deprecated and use fill instead */
```
jstnjns: Check this

src/styles/components/_badge.scss
```diff
@@ -1,11 +1,12 @@
-.badge {
+.mc-badge {
```
jstnjns: might want to verify that the previous class wasn't being used anywhere

src/styles/components/_tile.scss
```diff
@@ -149,11 +132,13 @@
 }
 
 .mc-tile-overlay {
+  transform: scale(1.01);
```
jstnjns: saw the "peeking background" happening again so i added this back in

stephenp: As long as it doesn't go back to having the same issue we were trying to fix where it bleeds out of the parent container...

src/assets/icons/play.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jstnjns: kill this, it's just like `ChevronRight` but without the proper viewbox spacing

src/styles/components/_tile.scss
stephenp: This comment piqued my interest so I took another look and fixed it.

src/assets/icons/play.svg
stephenp: Kept this because it's still 1:1 ratio. Don't want to bother design to export a new svg at the correct 24/24 since it would look the same anyway.

PR #357, opened 2019-01-30
modify Modal and VideoPlayer component to work with fullscreen NPS survey

a-axton:
## Overview
- Added `appendToBody` to Modal to disable/enable react portal appending to body
- Added fullscreen event listener to VideoPlayer to toggle `appendToBody` in chapter video NPS survey

## Risks
None

## Changes
No changes, all new props are optional

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -22,14 +22,12 @@ export default class Modal extends PureComponent {
     ]).isRequired,
     className: PropTypes.string,
     show: PropTypes.bool,
-    portalNode: PropTypes.instanceOf(Element),
-    usePortal: PropTypes.bool,
+    appendToBody: PropTypes.bool,
     onCloseClick: PropTypes.func,
   }
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/components/Modal/index.js
```diff
@@ -22,10 +22,16 @@ export default class Modal extends PureComponent {
     ]).isRequired,
     className: PropTypes.string,
     show: PropTypes.bool,
-
+    portalNode: PropTypes.node,
```
a-axton: oops wrong prop type, changing it now

src/components/VideoPlayer/index.js
```diff
@@ -131,6 +132,7 @@ export default class VideoPlayer extends PureComponent {
       }
     })
 
+    this.video.on('fullscreenchange', this.handleFullscreenChange)
```
bernabe9: In the `componentWillUnmout` method, add the following line to clean the listener:

```javascript
this.video.off('fullscreenchange')
```

PR #359, opened 2019-02-05
Carousel `peek` prop, improved documentation

jstnjns:
## Overview
Peek functionality, a design style for indicating there is more to see without controls (mobile requires a better use of real estate).

NOTE: This prop applies this style across the board.  If you want to apply this only to mobile, you'll need to use with the `ResponsiveHandler`.

## Risks
None

## Changes
screen shot 2019-02-05 at 1 03 05 am


src/components/TileOverlay/index.js
```diff
@@ -44,7 +44,7 @@ export default class TileOverlay extends PureComponent {
             rgba(${color}, 1) 0,
             rgba(${color}, 0.6) 25%,
             rgba(${color}, 0) 45%
-          )
+          ) center no-repeat
```
stephenp: Ayyy

PR #362, opened 2019-02-11
Fix modal background scroll

bernabe9:
## Overview
The body was scrolling when the modal was opened.

## Risks
Low. We will need to test mobile devices.


stephenp:
I remember when we first built out the modal, design really hated how you jumped to the top of the page when the modal opened, which is why we hadn't implemented the `position: fixed` solution initially.

I'm fine with this going in for now as long as it won't break anything (it doesn't look like it will) - and design signs off on it. I know we're getting ready to refactor modals in the next few days, so we might hold off on merging this since the whole thing will be redone shortly anyway. If this is blocking you from a feature, we can merge it in now.

PR #367, opened 2019-02-15
Input: Allow array of errors

bernabe9:
## Overview
The endpoints and the library [validate](https://validatejs.org/) that we use return an array of errors instead of returning just one error.

This PR adds support for array of errors in the inputs.

## Risks
None. String error still works

## Notes
@stephenp this PR is not urgent, but the log in & sign up modals are throwing prop-types errors (only visible on dev)


src/components/InputField/index.js
```diff
@@ -12,9 +12,18 @@ const INPUT_PROP_TYPE = PropTypes.shape({
...
+    PropTypes.arrayOf(PropTypes.string),
+  ]),
 })
 
+const parseError = (error) => {
+  if (!Array.isArray(error)) {
+    return error
+  }
+  return error.length > 0 ? error[0] : undefined
+}
```
jstnjns: i think i'd rather prefer the array logic to be in the conditional, and then anything else is "pass through as usual"

```
const parseError = (error) => {
  if (Array.isArray(error)) {
    return error[0]
  }
  return error
}
```

bernabe9: makes sense

jstnjns:
will make the requested changes in a further commit, this works

PR #374, opened 2019-02-26
Separator component

bernabe9:
## Overview
Simple component that renders a `<hr>` element to define a change in the content

## Risks
None. It's a new component

## Changes
screen shot 2019-02-26 at 4 35 36 pm


bernabe9:
I like the component because:
1. it forces the devs to use the correct html tag `<hr />`
2. it adds a semantic meaning to the element
3. We can still use just the styles and not the component

What do you think?

jstnjns:
yea, i'm good with that.  i have been thinking about a similar approach to type (like React Native `<Text>` with props for modifiers), but will wait a bit to implement something like that

PR #378, opened 2019-02-27
making sure the html font size stays as we set it

jstnjns:
## Overview
All major measurements (anything using the step scale) is based in REM, and therefore `html { font-size: ____; }`.  We are forcing the value of this so that we can expect consistent sizing throughout the library.

## Risks
Low - if our application is expecting to be able to set the font-size of the `html` attribute anywhere, this will override it.  However, we are globally loading mc-components, and anywhere the app might be doing this would cause rendering issues anyway.

## Changes
Stuff ALWAYS renders properly


bernabe9:
I tested locally and I found that now its displaying the font too small :(

 
screen shot 2019-02-27 at 12 59 15 pm

bernabe9:
@jstnjns pushed a commit with the body font size fix

Issue #385, opened 2019-03-06
Modal backdrop allows click through


fsarachu:
# Problem
Because of the way `ClickOutside` handle events, it's possible to click a button that is behind the modal backdrop.
Currently, to prevent this, we found ourselves placing the following snippet on the `onClose` handler to work it around:
```javascript
  onModalClose = (source, event) => {
    event.preventDefault()
    event.stopPropagation()
    // ......
    // The actual handling code
  }
```

# Possible Solution
We could handle this on the actual `Modal` component, calling `event.preventDefault()` and `event.stopPropagation()` if the click source is `backdrop`

bernabe9:
@fsarachu @jstnjns can we close this issue?

PR #386, opened 2019-03-06
Prevent Modal backdrop click through

fsarachu:
## Overview
Call stopPropagation() and preventDefault() if modal close source is 'backdrop'

## Risks
Low - Already tested on existing `Modal` usages, but more testing is always welcome

### Before:
before

### After:
after

## Issue
#385


src/components/Modal/index.js
```diff
@@ -49,6 +49,11 @@ export default class Modal extends PureComponent {
       onClose,
     } = this.props
 
+    if (source === 'backdrop') {
```
jstnjns: is there a reason this should only happen with the backdrop click?

fsarachu: Actually I added that because I wasn't 100% sure if that would have impact on other use cases...
Would it be safe to remove that `if`?

jstnjns: let's go ahead and remove it, and if we find any issues later we can fix them.  but i don't see why there would be any and prefer to not protect ourselves from something we aren't even sure exists

PR #418, opened 2019-03-25
Accordion: change hover styles so space-filling children don't trigger hover states

tamarajordan:
## Overview
Don't trigger hover state for all accordion children, since we insert some space-filling "dummy" children if we have fewer items to show than the number of accordion tiles desired for a given breakpoint (eg we want to show 4, but only have 3 items). Instead, set state when you hover over an actual item that needs a hover state.

## Risks
Low.. we're not using this accordion anywhere currently in the MC repo

## Changes
Affects hover state on accordion items, only dims if you are hovering over an item with content (not a space-flling item)

## Issue


src/components/Accordion/index.js
```diff
@@ -82,7 +88,20 @@ export default class Accordion extends PureComponent {
                     <HoverHandler nowrap>
                       {({ [on]: itemActive, props: itemProps }) =>
                         <div className={itemClasses(itemActive)} {...itemProps}>
-                          {renderChildren(child, { itemActive, parentActive })}
+                          <div
+                            className='mc-accordion__item-wrapper'
+                            onMouseEnter={() => {
```
jstnjns: could we move these to class methods?

src/playground/lihp/index.stories.js
```diff
@@ -266,7 +266,7 @@ storiesOf('Playground|Pages', module)
           <Accordion
             className='mc-mb-10'
             aspectRatio={responsiveValues(media, '21x9', '16x9', '4x3')}
-            on='hovering'
+            on='intent'
```
jstnjns: how does this look with intent since we are applying the styles immediately?

jstnjns: solved, works with whatever prop is passed, listening to both

PR #426, opened 2019-03-27
zero padding option for step, and TileCaption

jstnjns:
## Overview
Can now use the scale to set a zero value.  This adds helper classes like `.mc-m-0` and the like to explicitly set zero margin/padding.  This also means that `padding={0}` works for TileCaption.

## Risks
None, addition not change.


src/styles/base/_variables.scss
```diff
@@ -148,3 +148,7 @@ $btn-kinds: (
 
 // Forms
 $input-padding: 12px 16px !default;
+
+// Spacing
+$space-helper-beginning: 0 !default;
```
stephenp: Change to something like `$scale-from` and `$scale-to`

PR #427, opened 2019-03-27
bare link

jstnjns:
## Overview
Adding a link style we are commonly seeing in designs.  It's the exact same style as a `.mc-text--link` but only triggers the style if you hover over it.

There are also some undocumented classes for being able to apply this "reveal" on the hover of a much larger target by applying `.mc-text--bare-link-parent`.

## Risks
None

## Changes
Default
Screen Shot 2019-03-27 at 4 01 26 PM
Hover
Screen Shot 2019-03-27 at 4 01 31 PM


src/foundation/typography/modifiers.stories.js
```diff
@@ -25,15 +25,15 @@ const Modifiers = () =>
...
+      <PropExample name='.mc-text--link / .mc-text--bare-link'>
         <p className='mc-text-large'>
           The quick
           {' '}
           <a className='mc-text--link'>brown fox</a>
           {' '}
           jumped over the
           {' '}
-          <a className='mc-text--naked-link'>lazy dog</a>.
+          <a className='mc-text--bare-link'>lazy dog</a>.
```
stephenp: 😊

PR #442, opened 2019-04-01
TileVideo improvements (ref, positioning)

jstnjns:
## Overview
`TileVideo` needed to expose access to the `ref` of `VideoPlayer`.  It was also hardcoding account information.  If the `muted` prop changed, it would not pass that change to the `VideoPlayer` component.  The player is also positioned "center top" when it should be "center center" of the tile.  These are all updated in this request.

## Risks
Low - tested changes on both use cases in `masterclass`


src/components/TileVideo/index.js
```diff
@@ -6,19 +6,18 @@ import VideoPlayer from '../VideoPlayer'
...
     children: PropTypes.element,
-    controls: PropTypes.bool,
     className: PropTypes.string,
+    controls: PropTypes.bool,
     loop: PropTypes.bool,
     muted: PropTypes.bool,
-    videoId: PropTypes.string.isRequired,
+    onVideoReady: PropTypes.func,
+    playerRef: PropTypes.obj,
+    videoId: PropTypes.string,
```
stephenp: No longer `required`?

src/components/TileVideo/index.js
```diff
@@ -6,19 +6,18 @@ import VideoPlayer from '../VideoPlayer'
 import Unmute from '../Icons/Muted'
 import Mute from '../Icons/Unmuted'
 
-const ACCOUNT_ID = '5344802162001'
```
stephenp: Safe to remove because of the previous video player changes, right?

jstnjns: safe to remove because the video player already sets these values by default

src/components/TileVideo/index.js
jstnjns: that was a mistake

PR #446, opened 2019-04-03
Carousel - Increment Improvements

jstnjns:
## Overview
Prevents a Carousel from showing empty spaces.  It does this by hijacking the arrow click and calculating the last index is should scroll to.

## Risks
Low - has been tested on Storybook, but should be checked on masterclass as well

## Changes
No visual changes other than there will never be empty spaces on slides with multiple pages.


src/components/Carousel/index.js
```diff
@@ -75,26 +75,63 @@ export default class Carousel extends PureComponent {
     highlightOnHover: false,
     loop: false,
     overflow: false,
-    scrollCount: 1,
+    scrollCount: 2,
```
stephenp: Moving 2 at a time now?

src/components/Carousel/index.js
```diff
@@ -75,26 +75,63 @@ export default class Carousel extends PureComponent {
...
     showCount: 3,
     transition: TRANSITION_SLIDE,
   }
 
+  state = {
+    currentSlide: 0,
```
stephenp: Will this override the prop that lets you pass in what slide you want to start on? I know we've passed in custom slide starting values in the past

src/components/Carousel/index.js
```diff
@@ -75,26 +75,63 @@ export default class Carousel extends PureComponent {
...
+    super(props)
+
+    this.slider = props.sliderRef || React.createRef()
+  }
+
+  handlePrevClick = () => {
+    const { scrollCount } = this.props
+    const { currentSlide } = this.state
+
+    const goTo = currentSlide - scrollCount < 0
```
stephenp: ~goTo~ `targetSlide`?  And since we're handling previous click, would it make more sense to do `-1` vs doing the math to subtract `scrollCount` from `currentSlide`? Took me a while to figure this out

src/components/Carousel/index.js
```diff
@@ -75,26 +75,63 @@ export default class Carousel extends PureComponent {
...
+      ? 0
+      : currentSlide - scrollCount
+
+    this.slider.current.slickGoTo(goTo)
+  }
+
+  handleNextClick = () => {
```
stephenp: Same here, would it be simpler / more readable to just `+1` to the `currentSlide` value?

src/components/Carousel/index.js
```diff
@@ -75,26 +75,63 @@ export default class Carousel extends PureComponent {
...
       controls,
+      focusOnSelect,
       highlightOnActive,
       highlightOnHover,
       loop,
       overflow,
       peek,
       scrollCount,
       showCount,
-      sliderRef,
```
stephenp: Do we need to keep `sliderRef` here to handle multiple carousels controlling one another?

src/components/Carousel/index.js
```diff
@@ -16,27 +16,27 @@ class Arrow extends PureComponent {
     children: PropTypes.element,
     className: PropTypes.string,
     direction: PropTypes.oneOf(['left', 'right']).isRequired,
```
stephenp: Should this be updated to `prev` and `next` to match the nomenclature used below for `handlePress`?

src/components/Carousel/index.js
stephenp: Oh looks like it was changed from `sliderRef` to just `this.slider`, maybe this isn't needed then

src/components/Carousel/index.js
jstnjns: 👍

src/components/Carousel/index.js
jstnjns: yes, but only by default.  this is likely to be overridden

src/components/Carousel/index.js
jstnjns: good catch

src/components/Carousel/index.js
jstnjns: the value is not the increment, but the actual index to slide to

src/components/Carousel/index.js
jstnjns: the number to increment is a prop and can be changed

src/components/Carousel/index.js
stephenp: Heads up - I just did a quick search, but it does look like the `Carousel` is called at least one time in the main masterclass repo without specifying `scrollCount`. If we're ok with it updating to default to 2 everywhere and that behavior changing, disregard, but otherwise, that might need a find / replace before merging this in.

https://github.com/yankaindustries/masterclass/blob/staging/app/javascript/components/AllAccessPassCarousel/index.js#L30

src/components/Carousel/index.js
jstnjns: correct.  the ref is stored on the instance in the constructor because there's some logic associated to generating it now.  basically, the arrows need to be able to access the ref, but if one is not passed as a prop, there isn't one.  so we say use the one given or create one.

Issue #107, opened 2018-07-23
VideoPlayer doesn't autoplay on mobile (Safari)


bernabe9:
Why?
The following props are missing in the video HTML element: `muted` and `playsinline`

How to fix it?
We are using React to render the video HTML element, but that's not the recommended apporach by brightcove. Instead, we should inject the video HTML element using javascript.

Example: https://github.com/yankaindustries/mc-components/blob/55b3b2335378957e3f5bd7f870f0d9e3b561472c/src/utils/brightcove.js

stephenp:
Do you have an example of where videoPlayer is broken on mobile?

bernabe9:
I update the title to: VideoPlayer doesn't autoplay on mobile (Safari)

Yes, I have the example of the lesson plan header:
https://github.com/yankaindustries/masterclass/blob/staging/app/javascript/packs/client/components/lessonPlan/LessonPlanHeader/LessonPlanHeader.js#L41

If you remove my quick fix is not going to work.

The lesson plan header is on the homepage screen. To enable the experiment: `?lesson_plan=variation&forced=true`

jstnjns:
according to brightcove, it's better we don't ask safari to autoplay if there is audio https://support.brightcove.com/autoplay-considerations

if the implementation specifically asks for a muted video, we'll allow safari to autoplay.  otherwise we will not try to do anything with the video until the user interacts with it.

PR #456, opened 2019-04-10
allowing autoplay for safari if the video is muted

jstnjns:
## Overview
safari has it's own restrictions on autoplaying videos if the video has audio.  however, safari allows videos that are muted to play, so we are updating our logic accordingly.  eventually this logic will be abstracted to autoplay feature detection, but until now this is a simple implementation

## Risks
None

## Issue
#107


jstnjns:
@tamarajordan for sure.  i'll let you know when a PR on masterclass is available

PR #476, opened 2019-05-01
Adding a rounded class helper and symmetrical icon styles

stephenp:
## Overview
Adds a "symmetrical" and "rounded" property to the button component:
image

The rounded property can also be added to any image or block element with `.mc-corners--circle`

## Risks
None

## Changes
- Updated buttons to have "symmetrical" padding when that property is set
- Added new helper class to set any block element to have a full border radius (maybe avatars in the future?)

## Issue
N/A

### DIFF
--- a/src/styles/helpers/_corners.scss
+++ b/src/styles/helpers/_corners.scss
@@ -1,6 +1,6 @@
 .mc-corners {
   &--rounded {
-    border-radius: 4px;
+    border-radius: $default-radius;
   }
 
   &--circle {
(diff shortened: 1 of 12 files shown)


src/styles/helpers/_corners.scss
```diff
@@ -0,0 +1,13 @@
+.mc-corners {
+  &--rounded {
+    border-radius: 4px;
```
jstnjns: can we make this a variable?

PR #482, opened 2019-05-14
Fixing double toggle issue, better scroll styles for dropdown

jstnjns:
## Overview
Combo of being able to use the same click outside of a dropdown to open a new dropdown (or any other click handler), as well as toggle the dropdown.

## Risks
None - fixes bug

### DIFF
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -28,6 +28,7 @@ export default class Dropdown extends PureComponent {
 
   state = {
     show: false,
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/index.js
```diff
@@ -28,6 +28,7 @@ export default class Dropdown extends PureComponent {
 
   state = {
     show: false,
+    // lastTimestamp: 0,
```
stephenp: Remove

PR #492, opened 2019-05-22
Adding the ability for the Button to render another component/element

jstnjns:
## Overview
Adding the ability to have the Button choose a component or element to render (defaults to `button`).

## Risks
Low - There shouldn't be any unintended circumstances that I'm aware of

## Changes
Screen Shot 2019-05-21 at 5 00 07 PM
Screen Shot 2019-05-21 at 5 00 19 PM


jstnjns:
> Can you add this to the documentation somewhere?

I'm not sure I want to "expose" this just yet, but will def add to docs once we nail down it's usage.

PR #497, opened 2019-05-29
Dropdown - Controlled

jstnjns:
## Overview
Adding controlled Dropdown functionality

## Risks
None - new component and added functionality.  Nothing changing from existing implementations.

## Changes
Screen Shot 2019-05-29 at 11 24 53 AM

### DIFF
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -6,7 +6,6 @@ import Popper from 'popper.js'
 
 const DropdownContext = React.createContext('dropdown')
 
-
 export const {
   Provider,
   Consumer,
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -28,16 +28,19 @@ export default class Dropdown extends PureComponent {
     lastTimestamp: 0,
   }
 
+  toggleRef = React.createRef()
   dropdownRef = React.createRef()
+  arrowRef = React.createRef()
+
+  componentDidMount () {
+    this.renderDropdown()
+  }
 
   componentWillUnmount () {
     document.body.classList.remove('mc-dropdown__body--open')
   }
 
   toggle = (event) => {
-    const {
-      placement,
-    } = this.props
     const {
       lastTimeStamp,
       show,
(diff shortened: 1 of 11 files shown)


src/components/Tooltip/index.js
```diff
@@ -93,15 +93,6 @@ export default class Tooltip extends PureComponent {
     return data
   }
 
-  applyStyle = (data) => {
```
a-axton: was this supposed to be copied and not deleted?

src/components/Dropdown/index.js
```diff
@@ -28,16 +27,25 @@ export default class Dropdown extends PureComponent {
...
+    this.renderDropdown()
+  }
 
   componentWillUnmount () {
     document.body.classList.remove('mc-dropdown__body--open')
   }
 
+  componentDidUpdate (prevProps) {
+    if (prevProps.placement !== this.props.placement) {
+      this.renderDropdown()
```
a-axton: shouldn't we be destroying the old popper instance before creating a new one?

src/components/Tooltip/index.js
```diff
@@ -51,8 +51,8 @@ export default class Tooltip extends PureComponent {
     this.renderTooltip()
   }
 
-  componentWillReceiveProps (newProps) {
-    if (newProps.placement !== this.props.placement) {
+  componentDidUpdate (prevProps) {
+    if (prevProps.placement !== this.props.placement) {
       this.renderTooltip()
```
a-axton: same with dropdown, should we destroy the previous popper?

src/components/Tooltip/index.js
jstnjns: somehow it was a duplicate of https://github.com/yankaindustries/mc-components/pull/497/files/a705c179d6acce7c7b8ae7c9b049abf7a5457d66#diff-0ad95356130ae09835069df6f53fbdbeL87

src/components/Dropdown/index.js
jstnjns: i gave that a shot, but their destroy method seemed to just cause nothing to render anymore.  since placement almost never changes (really only in the docs) i wasn't super worried about doing my own cleanup since i presume the garbage collection will take care of abandoned nodes.

jstnjns: that, or we could just not support placement change

Issue #229, opened 2018-10-09
Pull `Header` and `Footer` out of mc-components, move to masterclass shared


jstnjns:
These are not components that make sense in a publicly shared library, they make more sense in the scope of the masterclass project.  This also makes it easier to perform experiments.

stephenp:
Only pulled out header, still need to remove footer.

stephenp:
Still have to remove these:

https://github.com/yankaindustries/mc-components/blob/develop/src/styles/components/_navbar-menu.scss

https://github.com/yankaindustries/mc-components/blob/develop/src/styles/release-candidate/components/_header.scss

stephenp:
Removed all the things

PR #501, opened 2019-06-10
Removing the rest of the things

stephenp:
## Overview
Missed a few remnants, let's get rid of them.

## Risks
Low

## Changes
No visual changes.

## Issue
https://github.com/yankaindustries/mc-components/issues/229


jstnjns:
kill it with 🔥

PR #503, opened 2019-06-11
Adding small buttons, and more interactive button documentation

jstnjns:
## Overview
Adding small modifier to Buttons
Also added a button "generator" so that you can preview different button styles

## Risks
None - adding a modifier, doesn't change any existing

## Changes
 
Screen Shot 2019-06-11 at 2 25 33 PM

## Issue
https://masterclass-dev.atlassian.net/browse/DS-122


src/components/Button/index.js
```diff
@@ -52,26 +57,22 @@ export default class Button extends PureComponent {
       className,
       fullWidth,
       kind,
-      link,
       loading,
       rounded,
-      secondary,
```
stephenp: Blaming @jstnjns when we forget to update these in the main masterclass repo before versioning :X

jstnjns: i'll allow it

src/styles/components/buttons/_buttons.scss
```diff
@@ -56,6 +63,7 @@
 
     .c-button__content {
       opacity: 0;
+      display: inherit;
```
stephenp: Y u need this?

src/styles/components/buttons/_buttons.scss
```diff
@@ -1,8 +1,8 @@
 @mixin mc-button {
   @include step(padding-top, 3);
-  @include step(padding-right, 5);
   @include step(padding-bottom, 3);
   @include step(padding-left, 5);
+  @include step(padding-right, 5);
```
stephenp: Wrong spot

src/styles/components/buttons/_buttons.scss
jstnjns: this fixes that weird `0.94px` issue when transitioning from an icon to a loading indicator

jstnjns: don't ask me why, but it does

stephenp: Can revisit in the future - think it might have something to do with the svg itself and the viewbox maybe? approving.

jstnjns: sure.  the odd part is that without this wrapper, it's fine.

text -> spinner === fine
icon by itself === fine
icon -> spinner !== fine

PR #513, opened 2019-06-20
Modal - Sizes

jstnjns:
## Overview
Instead of using the grid system, we've decided to have set widths for content sizes.  I've added demos for all four sizes (small, medium, large, full).

## Risks
High - the default size is set to the full width of the container.  Containers are currently in all implementations, so we'll need to remove those.  While this is high risk, the fallout is minimal (extra margin on the left and right of the modal on small screen sizes).

## Changes

Screen Shot 2019-06-19 at 9 44 47 PM
Screen Shot 2019-06-19 at 9 44 50 PM
Screen Shot 2019-06-19 at 9 45 29 PM
Screen Shot 2019-06-19 at 9 45 45 PM


stephenp:
https://github.com/yankaindustries/mc-components/issues/514

PR #523, opened 2019-07-19
Adding proper z-indexing for tooltips

stephenp:
## Overview
While @felipegarcia92 was paying back some tech debt on custom built tooltips (because mc-components didn't have tooltips yet!), he noticed that the tooltips were not showing up when used inside of a modal.  After some troubleshooting, we discovered that the tooltips were indeed rendering, but under the modal.

Because of how tooltips are moved to the body of the page and outside of the dom flow, they were being hidden behind the higher z-indexed modal.  We have a list of z-index variables in our variables file, so I used the tooltip variable which seems to resolve the issue:

```$mc-zindex-dropdown-backdrop:  900 !default;
$mc-zindex-dropdown:          1000 !default;
$mc-zindex-sticky:            1020 !default;
$mc-zindex-fixed:             1030 !default;
$mc-zindex-modal-backdrop:    1040 !default;
$mc-zindex-modal:             1050 !default;
$mc-zindex-popover:           1060 !default;
$mc-zindex-tooltip:           1070 !default;
```

## Risks
Low - there might be a case where we had some other items intentionally z-indexed higher than 1070 that the tooltip will now appear over while "active", but that's a pretty small risk.

## Changes
Tooltips will now render as one of the highest z-indexed items in mc-components.  If custom css is being used on a page, the tooltip z-index might not be high enough to appear above that content, in which case that content needs to be refactored.

## Issue
N/A

### DIFF
--- a/src/styles/components/_tooltip.scss
+++ b/src/styles/components/_tooltip.scss
@@ -8,8 +8,8 @@
 
   &--show {
     opacity: 1;
-
     pointer-events: auto;
+    z-index: $mc-zindex-tooltip;
   }
 
   &__toggle {


jstnjns:
we should also consider whether a dropdown should be able to exist in a modal, and if so raise that value

PR #530, opened 2019-08-02
Add modal class to the body when the modal mounts opened

### LINKED ISSUE #529: Modal - body class not added when mount & open the modal
mc-components' modal adds a class to the body to prevent background scrolling, but this class is added in the `componentDidUpdate` life cycle: https://github.com/yankaindustries/mc-components/blob/develop/src/components/Modal/index.js#L43

so, when you mount a modal already open it never add this class to the body.

bernabe9:
## Overview
Fix issue with the body class.

## Issue
Resolves #529

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -40,6 +40,14 @@ export default class Modal extends PureComponent {
     size: 'full',
   }
 
+  componentDidMount () {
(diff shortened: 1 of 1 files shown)


jstnjns:
nice, thanks!

PR #550, opened 2019-08-29
Fixing Toast, when rendered but `show` is set to false

jstnjns:
## Overview
If you had a toast without the `show` prop set, you'd see an intro animation and then it suddenly disappear.

## Risks
None - nothing is using this yet


src/components/Toast/index.js
```diff
@@ -35,14 +36,17 @@ export default class Button extends PureComponent {
     kind: 'default',
   }
 
-  toastRef = React.createRef()
+  containerRef = React.createRef()
```
jstnjns: Using an inner container instead of the parent.  This means that the height of the parent can be zero, but we can still calculate the height of the content (absolutely positioned)

src/components/Toast/index.js
```diff
@@ -35,14 +36,17 @@ export default class Button extends PureComponent {
...
 
   state = {
     show: false,
   }
 
   show = () => {
-    this.setState({ show: true })
+    this.setState({
+      height: this.containerRef.current.offsetHeight,
```
jstnjns: Setting the height here to ensure that the styles applied to the page already exist (was having issues with dynamically loaded styles, and the height calc being without any CSS)

src/components/Toast/index.js
```diff
@@ -84,29 +96,36 @@ export default class Button extends PureComponent {
...
-      show,
       height,
+      show,
     } = this.state
 
     const classNames = cn({
-      'mc-toast mc-mt-2 mc-py-1 mc-px-3': true,
-      'mc-toast--show': show,
+      'mc-toast': true,
+      'mc-toast--appear': show,
```
jstnjns: Same thing, different name.  This just keeps inline with the `CSSTransition` naming

PR #577, opened 2019-10-08
making a copy of ButtonLoader -> Loader

bcfancher-masterclass:
## Overview
Basically migrating ButtonLoader to something more generic, like Loader

## Risks
No Risk

## Changes
No change, new component

## Issue
fixes nothing - later I'll clean up masterclass experiments where folks are using ButtonLoader as a generic content loader (will change to use Loader instead)

## Breaking change?
backwards compatible


src/components/Loader/index.js
```diff
@@ -0,0 +1,91 @@
...
+import PropTypes from 'prop-types'
+
+const Loader = ({ color }) => (
+  <div className='loader'>
+    <svg
+      width='120'
+      height='30'
+      viewBox='0 0 120 30'
+      xmlns='http://www.w3.org/2000/svg'
+      fill={color}
```
jstnjns: I wonder if we might improve this later (not for this PR) by removing this forced color and using `'currentColor'`

bcfancher-masterclass:
> One thing we need to do when we add new components is have them exported from `src/components/index.js`. That's my bad for not sharing that or it being anywhere in the documentation.

Thanks, but I did know that.  Not sure that's worth documenting.  Good catch

Issue #207, opened 2018-09-14
Sticky component


stephenp:
Sticky nav, sticky header, sticky banner, etc.

- Offset from top
- Sticky from beginning, or sticky at scroll offset?

jstnjns:
Possible implementation: https://github.com/jstnjns/masterclass/blob/react-homepage/app/javascript/components/Fixed/index.js

jstnjns:
Actually, to keep this simple, let's just keep the props to: `position: ['top', 'bottom']`

bcfancher-mc:
both of the above components have been added to satiate these requirements

Issue #578, opened 2019-10-08
ScrollHandler component


jstnjns:
A component that, similar to HoverHandler, passes a prop to the children of whether you've scrolled to a certain pixel value down the page.  I think we can keep the prop of where you're scrolling "to" ambiguous...by calling the prop `to` 😂 

```jsx
<ScrollHandler to={600}>
  {({ scrolled }) => 
    <div>{scrolled ? 'You got to 600px!' : 'Keep scrolling...'}</div>
  }
</ScrollHandler>

PR #581, opened 2019-10-08
New Sticky Component

bcfancher-masterclass:
## Overview
New Sticky component to be used when you need to fix a component at the top or bottom of it's parent-positioned element.

## Risks
None

## Changes
Does not affect existing components

## Issue
https://github.com/yankaindustries/mc-components/issues/207

## Breaking change?
Backwards Compatible


src/styles/components/_sticky.scss
```diff
@@ -0,0 +1,15 @@
+.mc-sticky {
```
stephenp: Will also most likely need a z-index so it can layer above content on the page that has z-indexes specified: https://github.com/yankaindustries/mc-components/blob/develop/src/styles/base/_variables.scss#L69

src/components/Sticky/index.stories.js
```diff
@@ -0,0 +1,102 @@
...
+      [name]: value,
+    })
+
+  render () {
+    const { position } = this.state
+    const containerStyle = {
+      paddingTop: '60px',
+      paddingBottom: '1000px',
+    }
+    const navbarStyle = {
```
stephenp: If you want, you can move the example styles into the "upside down" scss file as well - it's where we've put any styles that aren't needed for mc-components, but _are_ needed for the documentation in storybook.  We called it upside down when stranger things had just come out... :D 

https://github.com/yankaindustries/mc-components/blob/develop/.storybook/_upside-down.scss

src/components/Sticky/index.js
```diff
@@ -0,0 +1,33 @@
+import React, { PureComponent } from 'react'
```
stephenp: One of the main components of the original issue / feature request was being able to set an offset from the top.  We currently have a fixed header on body scroll, so we'd have to write additional logic on top of this sticky component to figure out the offset.  The original experiment's implementation looked at scroll / offsets which you can see here:

https://github.com/jstnjns/masterclass/blob/react-homepage/app/javascript/components/Fixed/index.js#L34

I think this is something we should support, but @jstnjns can weigh in on whether that's needed for this particular sticky component.

bcfancher-masterclass: @stephenp, so @jstnjns and I discussed the 'showAfter' property this morning and he feels this particular behavior should be encapsulated as a separate component.  I can create another PR for that component.  I can add another property to this PR to handle the offset

jstnjns: yea.  sry @stephenp didn't update you on this.  realized that it really was separate functionality, and that we shouldn't conflate the utility of this component with the utility of what should be another component

bcfancher-masterclass: @jstnjns @stephenp would we want to call such a component 'ShowAfter' ?

bcfancher-masterclass: or maybe 'RevealOnScroll' ?

jstnjns: https://github.com/yankaindustries/mc-components/issues/578

For ease, I think it's fine to start with a pixel value.  If we REALLY need to pull in a ref or something, we can improve that later.

PR #583, opened 2019-10-09
added new RevealOnScroll component

bcfancher-mc:
## Overview
Added a new component to handle revealing a component after scrolling X pixels

## Risks
None

## Changes
N/A

## Issue
N/A

## Breaking change?
Backwards compatible / N/A


jstnjns:
i think i'd prefer the approach defined here https://github.com/yankaindustries/mc-components/issues/578 as it gives us the same functionality, but more flexibility

bcfancher-mc:
> i think i'd prefer the approach defined here #578 as it gives us the same functionality, but more flexibility

@jstnjns so specifically, rather than handling the reveal in the new component, simply provide a ScrollOn utility and pass the reveal onto the children?

Issue #588, opened 2019-10-14
Carousel w/ `peek` + variable width tiles wraps to second line


stepchud:
Gif of the issue can be viewed here: https://ourmasterclass.slack.com/files/UBT6FTC0M/FPBQT2PJB/oct-11-2019_01-49-18.gif

PR #590, opened 2019-10-15
fixing carousel variableWidth bug

jstnjns:
## Overview
<General description>

## Risks
Low - this should have no affect on existing carousels, but fix a problem with new implementations.  _should..._

## Issue
https://github.com/yankaindustries/mc-components/issues/588

## Breaking change?
Backwards Compatible


src/styles/components/carousel/_base.scss
```diff
@@ -61,6 +40,29 @@ $scrollbar-max-width: 17px !default;
...
+  }
+
+  .slick-slide {
+    opacity: 0;
+    pointer-events: none;
+    transition: opacity 250ms ease;
+  }
+
+  // Align carousel items to left
+  .slick-track {
```
stephenp: I think this was needed for when we have a low number of slides in a carousel - like if you have 2 items in a carousel they're centered instead of aligned left.

Issue #495, opened 2019-05-23
How to handle cursor use cases on tile components


stephenp:
It seems like we're coming across instances where the `Tile` component should _not_ have the `cursor: pointer;` css style set.

https://ourmasterclass.slack.com/archives/CGH3BN06M/p1558644695014000

Should we:
- Make the default a pointer, but have an option for users to unset that?
- Leave it as it is, but have users write custom css to unset?
- Change to no cursor by default and have users add cursor pointers?

Open to discussion!

@jstnjns @tamarajordan

tamarajordan:
I like keeping the cursor set to pointer as a default, since I think generally tiles are and should be clickable. So I'd vote in favor of custom CSS or allowing people to unset the default cursor styling

jstnjns:
I wonder if it's best to attach the pointer "style" (i put it in quotes because i'm not sure i would even classify it as that, more like an interaction indicator) to attributes that actually need indication, like `[href]` or `[onClick]`. i feel like this approach would solve our needs without having to override styles based on a change in the default interaction

bcfancher-masterclass:
I def vote for only using pointer style when href or onClick props are set.  @jstnjns lemme know and I'll make it so

bcfancher-mc:
gah, just noticed we're not currently passing any onClick or href stuffs to tiles, only on the tile parent  >.<

bcfancher-mc:
necessary changes have been made in MC and MCC repos

PR #595, opened 2019-10-17
[DNM] adding pointer style if tile has onClick prop set

bcfancher-mc:
## Overview
adding a prop check for onClick, which will in turn add a pointer style to the Tile component if present

## Risks
Low: if merged before the mc refactor, some tiles may be missing pointer styles.  Should be no risk otherwise

## Changes
will remove pointer styles from any Tiles which do not have any onClick props or are not otherwise custom-styled to have pointer

## Issue
https://github.com/yankaindustries/mc-components/issues/495

## Breaking change?
May remove pointer styles from some tiles (see above).  Otherwise backwards compatible

### DIFF
--- a/src/components/Tile/index.js
+++ b/src/components/Tile/index.js
@@ -32,6 +32,7 @@ export default class Tile extends PureComponent {
       'mc-tile': true,
       [`mc-tile--${aspectRatio}`]: aspectRatio,
       'mc-tile--crop': crop,
+      'mc-tile--pointer': typeof restProps.onClick === 'function',
     })
 
     return (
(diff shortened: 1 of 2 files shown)


stephenp:
I think I'm ok with this - if you're not using react (and using the tile classes, for example) you'd most likely have custom css to add the pointer, or be wrapping all of it in an anchor already.  If you _are_ using react and don't have an onclick, that was probably not written correctly and we can accept that regression.  Thoughts @jstnjns ?

bcfancher-mc:
https://github.com/yankaindustries/masterclass/pull/11342
I went ahead and patched up the implementations where folks faked-the-funk by manually adding tile classnames.  This other PR needs to go out first, then this one.

Issue #458, opened 2019-04-10
Remove vertical gutters from grid in lieu of spacing helpers


jstnjns:

stephenp:
More detail:

We have a class modifier called `no-gutters-vertical`. This is necessary to reimplement the default bootstrap grid behavior of *not* having gutters separating each col.

We initially added padding custom around cols to accomplish the original design spec.  This issue is requesting a removal of that padding and restoring the default bootstrap grid behavior, but we'll need to go verify every use case of the grid and manually add padding where necessary.

PR #602, opened 2019-10-23
[DNM] removing default top/bottom gutters from grids

bcfancher-mc:
## Overview
Removing default top/bottom gutters/padding from grids. Adding new classes to handle cases where vertical-gutters are desired

Dependent on: https://github.com/yankaindustries/masterclass/pull/11421 being merged

## Risks
Medium - I'll have to go through the MC repo and add vertical-gutters classes probably in several places so as not to impact the desired layouts

## Changes
Before:
image

After:
image

## Issue
https://github.com/yankaindustries/mc-components/issues/458

## Breaking change?
It can be a breaking change - will require vertical-gutters classes to be add all over the MC layouts in order not to break stuffs


bcfancher-mc:
Ready for Review

src/components/Modal/index.stories.js
```diff
@@ -128,7 +128,7 @@ class ModalExample extends Component {
                             future purchases.
                           </p>
 
-                          <div className='row small-gutters no-gutters-vertical'>
+                          <div className='row small-gutters no-gutters'>
```
jstnjns: no-gutters also remove horizontal gutters, which i don't believe is the goal of the existing code

src/styles/libraries/bootstrap-overrides/_grid.scss
```diff
@@ -9,7 +9,7 @@
 }
 
 .row {
-  margin: $grid-gutter-width / -2;
+  margin: 0 $grid-gutter-width / -2;
```
jstnjns: i think we get this for free from bootstrap's base grid

src/styles/libraries/bootstrap-overrides/_grid.scss
```diff
@@ -19,7 +19,7 @@
 .col,
 [class^="col-"],
 [class*=" col-"] {
-  padding: $grid-gutter-width / 2;
+  padding: 0 $grid-gutter-width / 2;
```
jstnjns: same here, i believe we get this for free

src/styles/libraries/bootstrap-overrides/_grid.scss
```diff
@@ -74,13 +90,17 @@
   }
 }
 
-.no-gutters-vertical {
-  margin-top: 0;
-  margin-bottom: 0;
+.no-gutters-vertical-deprecated {
```
jstnjns: correct me if i'm wrong, but we should be able to actually kill this, since there is now nothing defining no vertical margin/padding then not defining this will have the same effect

src/styles/libraries/bootstrap-overrides/_grid.scss
bcfancher-mc: @jstnjns are you saying this style can be removed completely?  (or at least the margin?)

src/styles/libraries/bootstrap-overrides/_grid.scss
bcfancher-mc: same question

src/styles/libraries/bootstrap-overrides/_grid.scss
bcfancher-mc: well, you didn't want to have to nest the gutters-vertical-deprecated, so we made the selectors a little more lax / less specific, which makes the gutters-vertical-deprecated styles cascade to all the children.  So if you want to turn that back off again for subsequent children, then you have to use this new class

src/components/Modal/index.stories.js
bcfancher-mc: it's a valid point.  I'll check and see if it implodes if removed

PR #612, opened 2019-11-04
Pass props to the Loader Component

bernabe9:
## Overview
Pass props to the Loader Component

## Risks
None


bernabe9:
@jstnjns changing the size of the loader

PR #619, opened 2019-11-11
Passing all props from VideoPlayer onto video tag

felipegarcia92:
## Overview
As common practice, I added passing the props from wrapper component VideoPlayer onto video player tag.

## Risks
Low

## Changes
Just code-wise, any extra prop shouldn't bother with the video tag.

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -452,6 +452,7 @@ export default class VideoPlayer extends PureComponent {
             data-video-id={videoId}
             muted={isMuted}
             controls={hasControls}
+            {...this.props}
           />
         </div>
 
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -452,6 +452,7 @@ export default class VideoPlayer extends PureComponent {
             data-video-id={videoId}
             muted={isMuted}
             controls={hasControls}
+            {...this.props}
```
jstnjns: this will pass all props, but React doesn't like if you pass non-html attributes to html elements.  generally how i approach this is is by doing:

```
const {
  beforescreenComponent,
  pausescreenComponent,
  endscreenComponent,
  accountId,
  playerId,
  videoId,
  hasControls,
  isMuted,
  ...restProps
} = this.props


<video {...restProps} />
```

jstnjns: note that you'll need to reference every expected prop you want to pull out of `restProps` in order to not pass invalid attributes

felipegarcia92: @jstnjns you got it

PR #620, opened 2019-11-11
Mobile Safari bug

juanlacruz:
## Overview
Modals on safari are being cut off.

If you start scrolling down, and you try to open a modal, you'll see that it is cut off depending on your current position. 

## Risks
Low

## Changes
removed `position: fixed` on `mc-modal__body--open` class.

## Screenshots
Screen Shot 2019-11-11 at 4 29 01 PM

* Same modal as the previous one, but this time, it was opened from the top
Screen Shot 2019-11-11 at 4 30 05 PM

### DIFF
--- a/src/styles/components/_modal.scss
+++ b/src/styles/components/_modal.scss
@@ -83,7 +83,6 @@
   }
 
   &__body--open {
-    position: fixed;
     overflow: hidden;
     height: 100vh;
     width: 100vw;


src/styles/components/_modal.scss
```diff
@@ -83,7 +83,6 @@
   }
 
   &__body--open {
-    position: fixed;
```
bcfancher-mc: actually we need the position fixed, it's the overflow we can remove

bcfancher-mc: also, it seems this is not only impacting mobile safari, but also web and mobile chrome (check the bottom of the page when modal is open).  I'm not sure why we have overflow auto there originally, but I haven't had any problems with it removed on safari and chrome (gonna go run off and check FF now...)

bcfancher-mc: confirmed, overflow is safely removed on FF also

stephenp:
the fixed positioning was originally there to prevent a double scrollbar (I think) on pages where the modal content exceeds the height of the viewport.  Fixed makes the body fit within `100vh` so the only scrolling that would happen would be inside of the modal.  This might not affect this particular experiment / modal, but you might need to verify it doesn't break other modals with longer content.

src/styles/components/_modal.scss
stephenp: Fixed positioning was originally to prevent a double scrollbar (I think) on pages where the modal content exceeds the height of the viewport. Fixed positioning makes the body content fit within 100vh so the only scrolling that would happen would be inside of the modal. 

This might not affect this particular experiment / modal, but you should test / verify on modals with content that overflows the window since this modal style affects every modal.

bcfancher-mc:
I see.  We need the fixed to prevent the BG from scrolling... such a dilema!

stephenp:
This will be resolved once https://github.com/yankaindustries/mc-components/pull/629 is merged in

Issue #628, opened 2019-11-19
Modal is cutting/not displaying the content on mobile devices 


bernabe9:
Device: iphone 11
Browser: Safari
<img src="https://user-images.githubusercontent.com/18464392/69160589-a0754b80-0ac8-11ea-98f6-ea50e8006177.PNG" height="800" />

<img src="https://user-images.githubusercontent.com/18464392/69160591-a0754b80-0ac8-11ea-87f1-f3c5c9918011.PNG" height="800" />

------
Original issue: https://masterclass-dev.atlassian.net/browse/GROW-1731
Modal trailer sometimes is not displaying the content in the gift page.

PR #629, opened 2019-11-19
Modal body scroll & overflow refactor

bcfancher-mc:
## Overview
Refactoring Modal component to simplify the containers, make them a little more legible, and to clean up the body anti-scrolling and long, scrolling modals.

## Risks
Low - impacts MOST modals in mc (there are some plain-old bootstrap modals still)

## Changes
no intended behavioral differences, except contents are no longer cutoff

## Issue
https://github.com/yankaindustries/mc-components/issues/628

## Breaking change?
backwards compatible

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -114,7 +114,9 @@ export default class Modal extends PureComponent {
             kind={backdrop}
           />
           <div className='mc-modal__viewport'>
-            {children}
+            <div className='container'>
+              {children}
+            </div>
           </div>
         </div>
       </Provider>
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -53,10 +53,8 @@ export default class Modal extends PureComponent {
     const rootHtml = document.getElementsByTagName('html')[0]
 
     if (!prevProps.show && show) {
-      document.body.classList.add('mc-modal__body--open')
       rootHtml.classList.add('mc-modal__html--open')
     } else if (prevProps.show && !show) {
-      document.body.classList.remove('mc-modal__body--open')
       rootHtml.classList.remove('mc-modal__html--open')
     }
 
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -50,18 +50,18 @@ export default class Modal extends PureComponent {
 
   componentDidUpdate (prevProps) {
     const { show } = this.props
-
-    // a little hackery here to prevent scrolling/reset
+    const rootHtml = document.getElementsByTagName('html')[0]
 
     if (!prevProps.show && show) {
-      const scrollOffset = window.scrollY
-      document.body.style.top = `-${scrollOffset}px`
       document.body.classList.add('mc-modal__body--open')
+      rootHtml.classList.add('mc-modal__html--open')
     } else if (prevProps.show && !show) {
-      const scrollOffset = -parseInt(document.body.style.top, 10)
       document.body.classList.remove('mc-modal__body--open')
-      document.body.style.top = undefined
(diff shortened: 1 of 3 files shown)


src/components/Modal/index.js
```diff
@@ -50,18 +50,18 @@ export default class Modal extends PureComponent {
...
     const { show } = this.props
-
-    // a little hackery here to prevent scrolling/reset
+    const rootHtml = document.getElementsByTagName('html')[0]
 
     if (!prevProps.show && show) {
-      const scrollOffset = window.scrollY
-      document.body.style.top = `-${scrollOffset}px`
       document.body.classList.add('mc-modal__body--open')
+      rootHtml.classList.add('mc-modal__html--open')
```
stephenp: This here should only need the root.  From there, in the css we can target it like this if we need both overflows:

```
.mc-modal__body--open {
  overflow: hidden;
  body { overflow: hidden; }
}
```

src/components/Modal/index.stories.js
```diff
@@ -44,8 +45,8 @@ class ModalExample extends Component {
         />
 
         <DocSection title='Demo'>
-          <div className='row'>
-            <div className='col-md-3'>
+          <div className='row' style={{ height: '2000px' }}>
```
stephenp: I'm guessing this is so the page itself forces a scrollbar?  If so, can you move this to a standalone div after the `.row`?

stephenp: (Don't want people confused as to why it needs an inline style on the row if they inspect it)

src/styles/components/_modal.scss
```diff
@@ -32,38 +32,29 @@
...
+  &__viewport {
+    position: relative;
+    width: auto;
+    margin: 1.75rem 0;
+    z-index: $mc-zindex-modal-backdrop;
+
+    // vertically centered stuffs
+    display: flex;
+    align-items: center;
+    min-height: 95vh;
```
stephenp: Should this just be "height"?

src/components/Modal/index.js
bcfancher-mc: good catch

src/components/Modal/index.stories.js
bcfancher-mc: good point

src/styles/components/_modal.scss
bcfancher-mc: nope, has to be min-heigh, apparently

jstnjns:
love it 😍

jstnjns:
before we merge, can we do a sanity check with mc? link locally and check major usages?

bcfancher-mc:
> before we merge, can we do a sanity check with mc? link locally and check major usages?

I did this yesterday

jstnjns:
of course you did you wonderful man

jstnjns:
fill out the PR description for the records, but the code looks g2g!

src/components/Modal/index.js
```diff
@@ -50,18 +50,19 @@ export default class Modal extends PureComponent {
...
     } else if (prevProps.show && !show) {
-      const scrollOffset = -parseInt(document.body.style.top, 10)
-      document.body.classList.remove('mc-modal__body--open')
-      document.body.style.top = undefined
-      window.scrollTo(0, scrollOffset)
+      rootHtml.classList.remove('mc-modal__html--open')
+    }
+
+    if (show) {
+      const elem = this.container.current
```
stephenp: Thanks stack overflow!

bcfancher-mc: si senor!

PR #632, opened 2019-11-22
Added check in Modal & Toast to pop from FS element, if appendToBody

bcfancher-mc:
## Overview
Purpose of this PR is to make the Modal and Toast mc-components a little smarter with how they portal, specifically in the use case of any fullscreen browser behavior

## Risks
Medium.  This change can impact any and all Modals & Toasts used on the MC website

## Changes
It allows modals and toasts to be delivered/rendered when the browser is in Fullscreen mode

## Issue
https://masterclass-dev.atlassian.net/browse/RET-1217

## Breaking change?
Backwards compatible


src/components/Modal/index.js
```diff
@@ -32,12 +32,18 @@ export default class Modal extends PureComponent {
     ]),
     appendToBody: PropTypes.bool,
     onClose: PropTypes.func,
+    allowFullscreen: PropTypes.bool,
```
jstnjns: should be the inverse, opting out should look like `<Modal disableFullscreen />` instead of `<Modal allowFullscreen={false} />`

src/components/Modal/index.js
```diff
@@ -46,6 +52,28 @@ export default class Modal extends PureComponent {
...
+    document.addEventListener('fullscreenchange', this.checkFullscreenState)
+  }
+
+  componentWillUnmount () {
+    document.body.classList.remove('mc-modal__body--open')
+    window.removeEventListener('fullscreenchange', this.checkFullscreenState)
+  }
+
+  checkFullscreenState = () => {
+    const { allowFullscreen } = this.props
```
jstnjns: if we inverse the prop, this can simply become an early exit

```
if (this.props.disableFullscreen) return
```

src/components/Modal/index.js
bcfancher-mc: done

src/components/Modal/index.js
```diff
@@ -32,12 +32,18 @@ export default class Modal extends PureComponent {
...
     appendToBody: PropTypes.bool,
     onClose: PropTypes.func,
+    disableFullscreen: PropTypes.bool,
   }
 
   static defaultProps = {
     appendToBody: true,
     backdrop: 'dark',
     size: 'full',
+    disableFullscreen: false,
```
jstnjns: not necessary, but not something to hold this up

src/components/Modal/index.js
```diff
@@ -46,6 +52,43 @@ export default class Modal extends PureComponent {
...
+    if (this.props.disableFullscreen) return
+
+    document.removeEventListener('fullscreenchange', this.checkFullscreenState)
+    document.removeEventListener('mozfullscreenchange', this.checkFullscreenState)
+    document.removeEventListener('webkitfullscreenchange', this.checkFullscreenState)
+    document.removeEventListener('msfullscreenchange', this.checkFullscreenState)
+  }
+
+  checkFullscreenState = () => {
+    if (this.props.disableFullscreen) return
```
jstnjns: triple checking huh?  :trollface:

src/components/FullscreenHandler/index.js
```diff
@@ -0,0 +1,50 @@
...
+]
+
+export default class FullscreenHandler extends PureComponent {
+  static propTypes = {
+    onFullscreenChange: PropTypes.func,
+    children: PROP_TYPE_CHILDREN.isRequired,
+  }
+
+  state = {
+    fullscreenElement: undefined,
```
jstnjns: the default value for `document.fullscreenElement` when not fullscreen is `null`.  not that it matters a whole lot, but we should probably keep with that, as when you move from FS to normal, we will be setting this value to `null`

src/components/FullscreenHandler/index.js
```diff
@@ -0,0 +1,50 @@
...
+      || document.msFullscreenElement
+
+    this.setState({ fullscreenElement })
+  }
+
+  render () {
+    const { children } = this.props
+    const { fullscreenElement } = this.state
+
+    return renderChildren(children, { fullscreenElement })
```
jstnjns: i wonder if it might be beneficial to pass along another prop that is the fullscreenElement when in fullscreen, and document.body when not

src/components/Fullscreen/index.js
```diff
@@ -0,0 +1,83 @@
...
+      if (this.props.isFullscreen) {
+        this.enterFullscreen()
+      } else {
+        this.exitFullscreen()
+      }
+    }
+  }
+
+  enterFullscreen () {
+    console.log('fullscreenElement: ', this.fullscreenElement)
```
jstnjns: remove

src/components/FullscreenHandler/index.stories.js
```diff
@@ -0,0 +1,170 @@
...
+    this.setState(prevState => ({ isFullscreen: !prevState.isFullscreen }))
+  }
+
+  toggleModal = () => {
+    this.setState(prevState => ({ showModal: !prevState.showModal }))
+  }
+
+  render () {
+    const { isFullscreen, showModal, toasts } = this.state
+    console.log('toasts: ', toasts)
```
jstnjns: remove

src/components/FullscreenHandler/index.js
bcfancher-mc: heard that - what would you call that, though?  currentRootNode?

bcfancher-mc: the more I think about it though, feels a bit out-of-bounds

jstnjns: let's sit on that idea for a bit before moving on it.  but i was thinking something like `screenElement`

Issue #638, opened 2019-12-02
Dropdown is scrolling to top on modal


bcfancher-mc:
Currently, when scrolled down a page and you click on a dropdown while on mobile, the body in the BG is scrolled to top. Also, content can be cutoff, depending on where you are scrolled on the page.  

This is very similar problem that we resolved on modals recently:
https://github.com/yankaindustries/mc-components/pull/629

PR #639, opened 2019-12-03
Fix dropdown scrolling and cutoff on mobile

bcfancher-mc:
## Overview
Currently, when scrolled down a page and you click on a dropdown while on mobile, the body in the BG is scrolled to top. Also, content can be cutoff, depending on where you are scrolled on the page.

This is very similar problem that we resolved on modals recently:
https://github.com/yankaindustries/mc-components/pull/629

## Risks
Low - may impact behavior of existing dropdowns

## Changes

## Issue
https://github.com/yankaindustries/mc-components/issues/638
NOTE: there is still some funky scroll behavior in the Dropdown documentation story; however it hasn't been replicated outside of that page, yet.  THAT solution is TBD

## Breaking change?
Backwards Compatible


felipegarcia92:
how can I invest in this one? haha

PR #640, opened 2019-12-03
VideoPlayer - Autoplay Fix

jstnjns:
## Overview
Autoplay for inline videos on mobile wasn't working.  Essentially, muting wasn't working properly, which Mobile Safari requires in order to autoplay.

## Risks
Medium - Affects autoplay for all players

## Changes
n/a

## Breaking change?
Backwards Compatible

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -414,6 +414,13 @@ export default class VideoPlayer extends PureComponent {
       hasControls,
       isMuted,
       isLooped,
+
+      // Pull out all of the props we don't want to pass along to the <video />
+      onPlayerReady,
+      onVideoReady,
+      onEnd,
+      onTimeChange,
+      onSeek,
       ...restProps
     } = this.props
 
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -133,7 +133,7 @@ export default class VideoPlayer extends PureComponent {
   }
 
   handlePlayerReady = () => {
-    const { onPlayerReady } = this.props
+    const { hasAutoplay, isMuted, onPlayerReady } = this.props
 
     // eslint-disable-next-line
     this.setState({ videoRoot: this.video.el_ })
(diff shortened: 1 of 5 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -21,8 +21,6 @@ const FILL_NONE = null
 
 const CC_HIDDEN = 'hidden'
 
-const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
```
jstnjns: removing this as we shouldn't be doing any checking.  if you want the video to work on safari (mobile or otherwise) mute the dang thing and make it part of the experience

src/components/VideoPlayer/index.js
```diff
@@ -155,6 +153,33 @@ export default class VideoPlayer extends PureComponent {
     this.calculateFill()
   }
 
+  handleReady = () => {
+    const { hasAutoplay, isMuted, onVideoReady } = this.props
+
+    if (isMuted) {
+      this.video.muted(true)
```
jstnjns: while we are passing `muted={isMuted}` in the render, there is evidence it is being stripped out by react

https://github.com/facebook/react/issues/6544

so we are programmatically muting here

src/components/VideoPlayerPortalScreen/index.js
```diff
@@ -1,14 +1,14 @@
...
+import { node } from 'prop-types'
 
 import { PROP_TYPE_CHILDREN } from '../constants'
 
 
 export default class VideoPlayerPortalScreen extends Component {
   static propTypes = {
     children: PROP_TYPE_CHILDREN,
-    videoRoot: instanceOf(Element),
+    videoRoot: node,
```
jstnjns: in some situations, Element doesn't exist.  using `PropTypes.node` is better

src/components/VideoPlayerPortalScreen/index.js
```diff
@@ -1,14 +1,14 @@
...
+import { object } from 'prop-types'
 
 import { PROP_TYPE_CHILDREN } from '../constants'
 
 
 export default class VideoPlayerPortalScreen extends Component {
   static propTypes = {
     children: PROP_TYPE_CHILDREN,
-    videoRoot: instanceOf(Element),
+    videoRoot: object,
```
jstnjns: `Element` doesn't always exist (like in SSR).  this isn't great, but it's better than failing

PR #642, opened 2019-12-06
Move theme-css-vars from base to helpers

bcfancher-mc:
## Overview
Just prevents themes from getting re-defined when folks import the following into their MC stuffs:
@import '~mc-components/dist/styles/scss/base/base';

## Risks
Low

## Changes
No expected affect

## Issue

## Breaking change?
Backwards compatible


src/components/DropdownHeader/index.js
```diff
@@ -5,7 +5,7 @@ import cn from 'classnames'
 import { PROP_TYPE_CHILDREN } from '../constants'
 
 
-export default class DropdownFooter extends PureComponent {
+export default class DropdownHeader extends PureComponent {
```
bcfancher-mc: Not sure how this change got squeezed into this branch, but I can remove if desired

jstnjns: how dare you name it what it should be named.  and who is the fool that named this incorrectly in the first place???

oh, it me...

PR #646, opened 2019-12-17
First Unit Tests for MCC

bcfancher-mc:
## Overview
Adding first tests to MCC to cover Dropdown component, so that I can confidently write tests for MC SearchDropdown and CategoriesDropdown.  

## Risks
None

## Changes
just adds tests and coverage metrics
image

## Issue

## Breaking change?
<Is this a breaking change, or is it backwards compatible? | *BREAKING* / Backwards Compatible>


.gitignore
```diff
@@ -19,3 +20,4 @@ package-lock.json
 # IDE Sharing
 .floo
 .flooignore
+.vscode
```
jstnjns: who uses vscode??

:trollface:

src/components/Dropdown/Dropdown.test.js
```diff
@@ -0,0 +1,268 @@
...
+    ) => (
+      <Dropdown placement={placement}>
+          <DropdownToggle><span>{toggleText}</span></DropdownToggle>
+          <DropdownContent>
+            <DropdownBody><span>{contentBody}</span></DropdownBody>
+          </DropdownContent>
+        </Dropdown>
+    )
+
+    const setup = () => {
```
jstnjns: i tend to prefer that each test does it's own setup explicitly so that render and setup functions don't get more complex with arguments being passed to components as props, as well as having to move further up the test file to see what setup and render functions are actually doing.  this is pretty simple at the moment, and dropdowns don't have a tendency to change a whole lot, but for testing practices as a whole i'm not convinced this is a pattern we want to reproduce.  i'm not going to request it's changed, though.  we can just keep an eye on this

.jest/setup.js
```diff
@@ -0,0 +1,10 @@
+import '@testing-library/jest-dom/extend-expect'
+
+window.matchMedia = window.matchMedia ||
+  function watchMedia () {
```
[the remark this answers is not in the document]
bcfancher-mc: great catch, I'll have to double-check what it's supposed to be...

src/components/Dropdown/Dropdown.test.js
bcfancher-mc: yeah, it's always a balance.  My approach usually is to group tests together as much as possible using contexts (or describes), and then create a setup method for each context.  This helps to prevent crazy setup methods w/ lots of args.

.gitignore
bcfancher-mc: 👀

bcfancher-mc: who uses floo??

.jest/setup.js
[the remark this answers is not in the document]
bcfancher-mc: removed altogether

Issue #653, opened 2020-01-14
Modal is scrolling the background


bernabe9:
![Jan-14-2020 17-03-14](https://user-images.githubusercontent.com/18464392/72378272-d06b5480-36ef-11ea-900e-ced7576c982d.gif)

PR #655, opened 2020-01-14
Fix modal background scroll

lunemmer:
## Overview
- The background was scrolling when the modal was opened: #653 
- If the modal has a button/link which redirects to another page, it keeps `mc-modal__html--open` class, making users unable to scroll in the new page

scroll

## Risks
Low 

## Changes
Affects mc-modals


jstnjns:
@lunemmer thanks for doing this?  was it tested on mobile as well?

lunemmer:
@jstnjns Yes and it's working!

I found in beta that you can scroll and "move" the modal. Is it expected behavior? I attached a gif so you can see that:
Jan-21-2020 12-33-32

PR #676, opened 2020-03-06
Adding catch for blocked autoplay

bcfancher-mc:
## Overview
there was an issue discovered where autoplay was throwing an exception.  This is as designed per browser autoplay policies

image

Google AutoPlay Policy Changes:
https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#new-behaviors

## Risks
None

## Changes
Before:
image

After:
image

## Issue
N/A

## Breaking change?
Backwards Compatible

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -165,7 +165,10 @@ export default class VideoPlayer extends PureComponent {
     }
 
     if (hasAutoplay) {
-      this.video.play()
+      const promise = this.video.play()
+      if (promise !== undefined) {
+        promise.catch(() => {})
+      }
     }
 
     this.checkBuffers()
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -165,7 +165,10 @@ export default class VideoPlayer extends PureComponent {
     }
 
     if (hasAutoplay) {
-      this.video.play()
+      const promise = this.video.play()
+      if (promise !== undefined) {
```
a-axton: why is this part needed? is it undefined under certain conditions?

bcfancher-mc: not sure, it's what was recommended on the google docs.

bcfancher-mc: I'd rather play it safe-than-sorry, given this is currently plaguing the packages player

jstnjns: word.  we can improve later if we find we want to

PR #693, opened 2020-04-08
New Color Prop for Carousels with Peek

bcfancher-mc:
## Overview
adding the ability to set the gradient color for peeking

## Risks
Low - default color is still black.  Should see no observable difference in behavior without this new prop.

## Changes
image

## Issue
N/A

## Breaking change?
Backwards Compatible

### DIFF
--- a/src/components/Carousel/index.js
+++ b/src/components/Carousel/index.js
@@ -76,6 +76,7 @@ export default class Carousel extends PureComponent {
     sliderRef: PropTypes.func,
     transition: PropTypes.string,
     variableWidth: PropTypes.bool,
+    color: PropTypes.string,
   }
 
   static defaultProps = {
@@ -91,6 +92,7 @@ export default class Carousel extends PureComponent {
     showCount: 3,
     transition: TRANSITION_SLIDE,
     variableWidth: false,
+    color: '0, 0, 0',
   }
 
   arrowContainerRef = React.createRef()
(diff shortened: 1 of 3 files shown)


src/components/Carousel/index.js
```diff
@@ -76,6 +76,7 @@ export default class Carousel extends PureComponent {
     sliderRef: PropTypes.func,
     transition: PropTypes.string,
     variableWidth: PropTypes.bool,
+    color: PropTypes.string,
```
stephenp, line 79: I think this is supposed to be `colorForThatFantasticFunkInPeekMode`

src/components/Carousel/index.js
```diff
@@ -167,6 +170,26 @@ export default class Carousel extends PureComponent {
       'mc-carousel__slider': true,
     })
 
+    let peekStyles
+    if (peek) {
+      peekStyles = {
```
stephenp, line 175: I think you'll only need to set the `background` styles here if everything else is already coming from the modifiers.scss file. We do this on the `TileOverlay` component: https://github.com/yankaindustries/mc-components/blob/master/src/components/TileOverlay/index.js

stephenp: Disregard - I remember the convo in slack now about not being able to style `:after` pseudo elements.

src/components/Carousel/index.js
bcfancher-mc: dammit!  good catch

PR #699, opened 2020-04-13
Fix LinkedIn icon with base SVG

jstnjns:
## Overview
Adding base SVG for LinkedIn icon.  This fixes an issue where rebuilding the icons directory removed the LinkedIn icon because there was no base SVG.

## Risks
None - fix

## Changes
n/a

## Breaking change?
Backwards Compatible


src/components/Icons/LinkedIn.js
```diff
@@ -1,9 +1,9 @@
...
+      d='M18.814 4H5.186C4.542 4 4 4.508 4 5.153v13.695C4 19.49 4.525 20 5.186 20h13.628c.66 0 1.186-.525 1.186-1.152V5.152C20 4.525 19.475 4 18.814 4zM8.767 9.979v7.633h-2.4V9.979h2.4zm-1.2-3.8c.771 0 1.388.61 1.388 1.374 0 .763-.617 1.374-1.388 1.374-.771 0-1.388-.61-1.388-1.374 0-.763.617-1.374 1.388-1.374zm4.929 3.828h-2.287v7.665h2.372V13.89c0-1 .186-1.966 1.423-1.966 1.22 0 1.236 1.136 1.236 2.017v3.714h2.372v-4.189c0-2.068-.44-3.645-2.846-3.645-1.152 0-1.914.644-2.236 1.238h-.034v-1.052z'
```
Alxzu, line 6: Not the same values as the one above?

jstnjns: SVGO compresses a little bit (the degrees of precision in the decimals it looks like).  I'm deploying so that we can see if the output is fine.

PR #702, opened 2020-04-16
Incident-128: Added fullscreen/descendent check to modal

bcfancher-mc:
## Overview
There was a problem discovered on production where videos within a modal could not enter into fullscreen mode.  This PR fixes that, while also still allowing modals to be opened within a fullscreen element.

#### Logged out
go to a cm page
click header video to open modal
click the fullscreen button and ensure it enters fullscreen mode successfully
#### Logged in
go to a class page
start last lesson in class
seek to end of video
ensure the NPS modal appears successfully

## Risks
None - this fixes an existing issue
## Changes

## Issue
Not sure if a ticket or issue was filed yet

## Breaking change?
Backwards compatible

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -141,23 +141,28 @@ export default class Modal extends PureComponent {
     return false
   }
 
+  getRootNode = (fullscreenElement) => {
+    let rootNode
+    if (fullscreenElement &&
+    !this.isDescendant(this.container.current, fullscreenElement)) {
+      rootNode = fullscreenElement
+    } else {
+      rootNode = document.body
+    }
+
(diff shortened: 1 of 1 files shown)


src/components/Modal/index.js
```diff
@@ -129,16 +129,35 @@ export default class Modal extends PureComponent {
     )
   }
 
+  isDescendant = (parent, child) => {
```
a-axton, line 132: would this `closest()` work instead of this? https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

bcfancher-mc: yep, worked beautifully, thanks!

PR #703, opened 2020-04-17
Display version in the docs

bernabe9:
## Overview
Display version in the docs to make sure that I'm reading the latest version of the docs

## Risks
None

## Changes
image

## Breaking change?
No


src/foundation/introduction/index.stories.js
```diff
@@ -5,6 +5,8 @@ import Highlight from 'react-highlight'
 import DocHeader from '../../utils/DocHeader'
 import DocSection from '../../utils/DocSection'
 
+import packageJSON from '../../../package.json'
```
stephenp, line 8: I never thought of doing this, but it makes complete sense. Great addition!

PR #704, opened 2020-04-17
Fix modal viewport height to support mobile devices

bernabe9:
## Overview
There's a known issue with `vh` units on mobile devices (https://css-tricks.com/the-trick-to-viewport-units-on-mobile/). Basically `100vh` doesn't represent 100% of the view height.

This PR fixes the viewport to support those devices. 

## Risks
Low. We are currently using this change in one feature tested on multiples devices.

## Changes
**Before (modal isn't centered):**

**After:**

## Breaking change?
No


stephenp:
So the browser chrome can vary in height depending on the browser you use and the phone OS you're on. Would it make sense to implement something like this instead?

https://github.com/mvasin/react-div-100vh

stephenp:
Merging this in, but created an issue for the library support mentioned: https://github.com/yankaindustries/mc-components/issues/707

PR #709, opened 2020-04-22
Hide video play button on beforeScreen

mdornel:
## Overview
A big video play button is shown on the before screen passed down as `beforeScreen` prop into `VideoPlayer` component, this behavior is unexpected.

This PR hides the video play button when `beforeScreen` is used.

## Risks
Low, just some style change.

## Changes
Before:
Screen Shot 2020-04-22 at 11 42 13

After:
Screen Shot 2020-04-22 at 11 29 34

## Breaking change?
<Is this a breaking change, or is it backwards compatible? | *BREAKING* / Backwards Compatible>

### DIFF
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -479,10 +479,12 @@ export default class VideoPlayer extends PureComponent {
     } = this.state
 
     const isScreenOpen = screen !== SCREEN_NONE
+    const isScreenBefore = screen === SCREEN_BEFORE
 
     const containerClasses = cn({
       'bc-player': true,
       'bc-player--screen-open': isScreenOpen,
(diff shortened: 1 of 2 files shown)


src/components/VideoPlayer/index.js
```diff
@@ -479,10 +479,12 @@ export default class VideoPlayer extends PureComponent {
     } = this.state
 
     const isScreenOpen = screen !== SCREEN_NONE
+    const isScreenBefore = screen === SCREEN_BEFORE
 
     const containerClasses = cn({
       'bc-player': true,
       'bc-player--screen-open': isScreenOpen,
+      'bc-player--beforescreen': isScreenBefore,
```
stephenp, line 487: Thoughts on this being updated to be something like `bc-player--beforescreen-open` to match the above line?

Issue #712, opened 2020-05-04
Add "tap/click to unmute" to VideoPlayer where hasAutoplay and isMuted


jstnjns:
![image](https://user-images.githubusercontent.com/249444/81019667-f8580280-8e1b-11ea-899d-be4f7850ba71.png)

PR #713, opened 2020-05-07
Adding "tap to unmute" for muted autoplay videos

jstnjns:
## Overview
For videos that autoplay in browsers that don't support autoplay with audio, we mute the video, add an additional control that calls out "tap to unmute" and then autoplay the video.

## Risks
Low - if there are other overlays on videos there could be potential visual conflicts

## Changes
image

Screen Shot 2020-05-06 at 10 47 01 PM

## Issue
#713 

## Breaking change?
Backwards Compatible


jstnjns:
#712

PR #716, opened 2020-05-11
Carousel - Improvements

jstnjns:
## Overview
In an effort to standardize carousel implementations, DS has decided to focus on three standard carousel combinations.  This is now reflected in the documentation, along with some style updates (like dots and peek).

## Risks
Medium - we use carousels a lot, with many different

## Changes
image

## Issue
https://masterclass-dev.atlassian.net/browse/DS-167

## Breaking change?
Backwards Compatible


jstnjns:
These changes have already been deployed to the documentation site, so you can review changes there.

src/styles/components/carousel/_base.scss
```diff
@@ -29,6 +33,10 @@ $scrollbar-max-width: 17px !default;
     position: relative;
   }
 
+  // &__slider-container {
```
bcfancher-mc, line 36: pew pew

src/components/Carousel/index.js
```diff
@@ -211,11 +178,10 @@ export default class Carousel extends PureComponent {
                 slidesToShow={variableWidth ? 1 : adjustedShowCount}
                 infinite={loop}
                 draggable={false}
-                afterChange={this.handleAfterChange}
-                onInit={() => this.setState({ initialized: true })}
                 variableWidth={variableWidth}
-                lazyLoad='ondemand'
```
a-axton, line 217: was removing this intentional?

bcfancher-mc:
is it me?  or do some of these storybook examples do nothing?
image

stephenp:
> is it me? or do some of these storybook examples do nothing?
> image

It's hard to only enable one prop in each example...you'll have to use the arrow keys after clicking "in" to that carousel to navigate left and right.  I think we should prob add arrows on every carousel for demo purposes....

src/components/Carousel/index.js
jstnjns: sure wasn't

jstnjns: ah, there was an issue with variablewidth and this working together.  but i forgot to add it back in conditionally

jstnjns:
> is it me? or do some of these storybook examples do nothing?
> image

updated to show two at this size, as well as add controls

PR #722, opened 2020-05-15
VideoPlayer - changing `videoId` obeys `hasAutoplay`

jstnjns:
## Overview
Adhere's to the original `hasAutoplay` prop functionality, even when you change videos.

### Before
`hasAutoplay` is off, the video does not autoplay.  you change the `videoId`, and the new video would automatically play

### After
`hasAutoplay` is off the video does not autoplay.  you change the `videoId`, and the new video does not autoplay as well

## Risks
Low - the autoplay on second video could be expected, but i do not believe change this would be considered breaking functionality

## Breaking change?
Backwards Compatible

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -38,8 +38,9 @@ module.exports = async ({ config, mode }) => {
     test: /\.s[ac]ss$/i,
     use: [
       'style-loader',
-      'css-loader',
-      'sass-loader'
+      { loader: 'css-loader', options: { importLoaders: 1 } },
+      'postcss-loader',
+      'sass-loader',
     ],
     include: path.resolve(__dirname, '../'),
   })
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -192,9 +192,9 @@ export default class VideoPlayer extends PureComponent {
 
   autoplay = () => {
     if (this.props.hasAutoplay) {
-      this.play()
+      this.unmuteAutoplay()
     } else {
-      this.pause()
+      this.video.pause()
     }
   }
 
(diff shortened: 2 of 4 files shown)


.storybook/webpack.config.js
```diff
@@ -38,8 +38,9 @@ module.exports = async ({ config, mode }) => {
     test: /\.s[ac]ss$/i,
     use: [
       'style-loader',
-      'css-loader',
-      'sass-loader'
+      { loader: 'css-loader', options: { importLoaders: 1 } },
+      'postcss-loader',
```
jstnjns, line 42: adding postcss-loader because we do use inline svgs in the styles for video player icons.  in the documentation, these weren't showing

src/components/VideoPlayer/index.js
```diff
@@ -100,10 +100,13 @@ export default class VideoPlayer extends PureComponent {
...
   componentDidUpdate (prevProps) {
-    if (this.props.videoId !== prevProps.videoId
-      || this.props.hasAutoplay !== prevProps.hasAutoplay) {
+    if (this.props.videoId !== prevProps.videoId) {
       this.replaceWith(this.props.videoId)
     }
+
+    if (this.props.hasAutoplay !== prevProps.hasAutoplay) {
+      this.autoPlay()
+    }
```
jstnjns: replace should only be called when you are in fact replacing the `videoId`, separating out the autoplay functionality if that has changed

src/components/VideoPlayer/index.js
```diff
@@ -100,10 +100,13 @@ export default class VideoPlayer extends PureComponent {
...
   componentDidUpdate (prevProps) {
-    if (this.props.videoId !== prevProps.videoId
-      || this.props.hasAutoplay !== prevProps.hasAutoplay) {
+    if (this.props.videoId !== prevProps.videoId) {
       this.replaceWith(this.props.videoId)
     }
+
+    if (this.props.hasAutoplay !== prevProps.hasAutoplay) {
+      this.play()
+    }
```
jstnjns, line 109: separating the actions of changing the videoId and hasAutoplay

Issue #719, opened 2020-05-12
Add apple button and icon


a-axton:
We're adding apple sign in and already have apple pay. We should add the button styles to mc-components (black background/white background solid black border)

stephenp:
https://github.com/yankaindustries/masterclass/pull/13254/files#r424003781

jstnjns:
this issue is partially resolved, missing the icon

PR #725, opened 2020-06-02
<Button kind='apple' />

jstnjns:
## Overview
Replacing the ApplePay button with Apple, as they should be unified and not so specific.  Apple's branding guidelines work for all of it's buttons.

## Risks
Low - this deprecates `<Button kind='apple-pay'/>` and replaces it with <Button kind='apple' />.  Those instances need to get replaced.

## Changes
Before
image

After
image

## Breaking change?
*BREAKING* - but not major


jstnjns:
#719

PR #727, opened 2020-06-03
Scale values to utilize CSS variables

jstnjns:
## Overview
Moving scale values to CSS variables. This allows us to reduce the LOC and complexity around using scale step values. The CSS output goes from 130kb -> 116kb (~5500 -> ~4500 LOC). Previously, we would use a SCSS mixin:

```scss
@import "~mc-components/dist/styles/scss/base/base";

.example {
  @include step(padding: 6);
}
```

which would output:

```css
.example {
  padding: 2rem;
}

@media (min-width: 768px) {
  .example {
    padding: 2.2rem;
  }
}

@media (min-width: 960px) {
  .example {
    padding: 2.4rem;
  }
}
```

Instead, we can now just use the CSS variable, and the our code is the same as the output

```scss
.example {
  padding: var(--mc-scale-6)
}
```

## Risks
Low - the existing mixin, which is used quite a bit, still works.  It just outputs the CSS variable now instead.

## Breaking change?
Backwards Compatible

### DIFF
--- a/src/styles/base/mixins/_step.scss
+++ b/src/styles/base/mixins/_step.scss
@@ -1,20 +1,5 @@
 @mixin step($attribute, $value, $i: false) {
   $important: if($i, "!important", "");
-  $calc-sm: calc-step($value, $mc-step-decay-sm);
-  $calc-md: calc-step($value, $mc-step-decay-md);
-  $calc-lg: calc-step($value, $mc-step-decay-lg);
 
-  #{$attribute}: $calc-sm * 1rem #{$important};
-
-  @if $calc-md != $calc-sm {
-    @media (min-width: $mc-bp-md) {
-      #{$attribute}: $calc-md * 1rem #{$important};
-    }
-  }
-
-  @if $calc-lg != $calc-sm {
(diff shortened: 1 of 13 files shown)


src/styles/base/mixins/_step.scss
```diff
@@ -1,20 +1,5 @@
...
-      #{$attribute}: $calc-md * 1rem #{$important};
-    }
-  }
-
-  @if $calc-lg != $calc-sm {
-    @media (min-width: $mc-bp-lg) {
-      #{$attribute}: $calc-lg * 1rem #{$important};
-    }
-  }
+  #{$attribute}: var(--mc-scale-#{$value}) #{$important};
```
stephenp, line 4: If this is being simplified, that means all existing implementations of the step helper will need to be updated to use css vars in the main masterclass repo, right?

jstnjns: this mixin is those instances. it's effectively retrofitting all of them with the new way so we don't have to replace them all

stephenp: duh.

PR #747, opened 2020-07-28
Fix end screen blocking clicks

bernabe9:
## Overview
After the end screen exits, the user wasn't able to interact with the video player clicking the screen for pausing or resuming.

Looks like a wrong state of the transition was adding a class to the end screen component when it should be hidden.
image (30)

More info here: https://masterclass-dev.atlassian.net/browse/WE-202

## Risks
Medium. Not sure why we had that function for defining when the transition ends, but maybe the code reviewers can clarify this.

## Breaking change?
No


jstnjns:
do screens properly transition out?  or just hide immediately?  i don't see a lot of value in them transitioning i suppose other than aesthetic

bernabe9:
@jstnjns I have added a timeout, so now the start screen and pause screen have a transition out

PR #777, opened 2020-11-13
Tags component

jstnjns:
## Overview
Adding new `<Tags />` component!

https://yankaindustries.github.io/mc-components/?path=/story/components-forms-tags--tags

image

It accepts two props: `collection` and `onChange`.  Collection is an array of strings to be rendered as badges, and `onChange` is a callback method for when something has occurred within `Tags` that passes back the modified collection.  An example:

```
const collection = ['a', 'b', 'c']

// User clicks "X" on 'b' Badge

onChange(['a', 'c'])
```

This is a controlled component, so you'll need to maintain state for it.  This is incredibly simple with `useState`

```
const Example = () => {
  const [items, setItems] = useState([])

  return (
    <Tags collection={items} onChange={setItems} />
  )
}
```

**NOTE:** A follow-up component will eventually be created to automate the connection for integrated forms (`react-final-form` and `redux-form`).

---

Also grouped in some changes to Badge that make the colors (other than Primary) make sense based on the theme.  Removed `kind="bright"` and replaced with `kind="dim"`.  The `default` color with be high contrast to whatever the background of the theme is, and `dim` will be low contrast.

image

## Risks
Low - `Tags` component is new, but there are some mild changes to the `Badge` component to improve size and padding when it has an `Icon`.

## Issue
n/a

## Breaking change?
Breaking - make sure to verify no existing badges change expected behaviors in the main repo

### DIFF
--- a/.storybook/_upside-down.scss
+++ b/.storybook/_upside-down.scss
@@ -24,6 +24,21 @@ pre {
   font-family: monospace;
 }
 
+kbd {
+  font-family: monospace;
+  display: inline-block;
+  padding: 0 0.3em;
+  background: $mc-color-gray-300;
+  border: 1px solid $mc-color-gray-400;
+  border-radius: $default-radius;
+
+  .mc-invert &,
+  .mc-theme-light & {
+    background: $mc-color-gray-600;
+    border: 1px solid $mc-color-gray-500;
+  }
+}
+
 a {
   color: inherit;
 }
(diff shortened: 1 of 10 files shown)


stephenp (approved):

stephenp:
Only concern I have is making sure that those badge changes aren't going to mess up in the main repo - should spot check them to make sure they look correct in context. Should flag in the PR description as possibly "breaking" change so that we pay attention to that when versioning mc-components. @jstnjns

lunemmer:
🚀 🚀 🚀

PR #782, opened 2020-12-01
Updating light dark helper ordering and labeling for clarity

stephenp:
## Overview
image

The inverted helper now shows our default (dark) first, then light, and labels them accordingly. This PR also lightly adjusts the default background color so that the dark / light theme colors show up a bit better.

## Risks
Low - documentation change only

## Changes
See the pic above

## Issue
N/A

## Breaking change?
Backwards Compatible


src/utils/InvertedMirror.js
```diff
@@ -29,23 +29,23 @@ export default class InvertedMirror extends PureComponent {
...
               {children}
             </Background>
           </div>
 
           <div className='col-md-6'>
             <Background
-              color='dark'
-              className='mc-card'
+              color='light'
+              className='mc-card mc-invert'
```
jstnjns, line 45: should prolly use `.mc-theme-light` here

```suggestion
              className='mc-card mc-theme-light'
```

jstnjns (approved):

PR #792, opened 2020-12-30
Adding placeholder prop to tag input

lunemmer:
## Overview
Adding placeholder prop to mc tag input.

## Risks
Low.

## Changes

With placeholder:
image

With placeholder and email added:
image

Without placeholder:
image

https://www.loom.com/share/92e9c032880549c598e005d7913bad71

## Breaking change?
Backwards Compatible


missx (dismissed):

src/components/Tags/index.js
```diff
@@ -100,12 +107,13 @@ const Tags = ({
       <input
         type='text'
         ref={inputRef}
-        className='mc-form-element__element'
+        className={inputClasses}
         onChange={handleInputChange}
         onFocus={handleInputFocus}
         onBlur={handleInputBlur}
         onKeyDown={handleKeyDown}
         size={value.length || 1}
```
jstnjns, line 108: i wonder if adding the placeholder length here instead of additional CSS would be better?

```suggestion
         size={value.length || placeholder.length || 1}
```

jstnjns (dismissed):
if that doesn't work, here's an approval anyway￼

src/components/Tags/index.js
lunemmer: It works! Done 😃

PR #795, opened 2021-01-05
CheckboxField to work with both redux-form and final-form

jstnjns:
## Overview
`CheckboxField` was not working correctly with `final-form` implementations. We'd have to handle the checked state manually instead of relying on the controlled aspect of `final-form.  The documentation for `CheckboxField` has also been updated to use `final-form`.

## Risks
Medium - existing implementations of `final-form` that use the `CheckboxField` should be verified, since overriding checked state is no longer necessary.  However, this should be backwards compatible.

## Issue

## Breaking change?
Backwards Compatible

### DIFF
--- a/src/components/CheckboxField/index.js
+++ b/src/components/CheckboxField/index.js
@@ -26,10 +26,10 @@ const CheckboxField = ({
 }) => (
   <FormGroup {...input} {...meta}>
     <Checkbox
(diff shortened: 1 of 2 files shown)


stephenp:
https://github.com/yankaindustries/masterclass/pull/15531

